PR #17, opened 2020-08-29
Add mean squared logarithmic loss function

AakashKumarNain:
1. Added mean squared logarithmic loss function
2. In the `CONTRIBUTIONS.md` file, there is no mention of `code-style`. So, I am using 4-spaces.
3. I haven't formatted the code using `black` as there is no mention of any formatter as well.

I will add the tests once the above points are clear


objax/functional/loss.py
```diff
@@ -48,6 +52,25 @@ def cross_entropy_logits_sparse(logits: JaxArray, labels: Union[JaxArray, int])
     return logsumexp(logits, axis=1) - logits[jn.arange(logits.shape[0]), labels]
 
 
+def mean_squared_logarithmic_error(y_true: JaxArray,
+                                    y_pred: JaxArray,
```
david-berthelot, line 56: We follow PEP8 style alignment (with 120 chars).

objax/functional/loss.py
```diff
@@ -48,6 +52,25 @@ def cross_entropy_logits_sparse(logits: JaxArray, labels: Union[JaxArray, int])
...
+                                    eps: float = 2 ** -17) -> JaxArray:
+    """Computes the the mean squared logarithmic error.
```
david-berthelot, line 58: Is there a link to a paper or to wikipedia we could add here?

objax/functional/loss.py
```diff
@@ -48,6 +52,25 @@ def cross_entropy_logits_sparse(logits: JaxArray, labels: Union[JaxArray, int])
...
+        y_pred: n-dimensional tensor of floats.
+
+    Returns:
+        scalar tensor containing mean squared logarithmic error
+    """
+    y_true_log = jn.log1p(jn.maximum(y_true, eps + 1.))
+    y_pred_log = jn.log1p(jn.maximum(y_pred, eps + 1.))
+    
+    loss = (y_pred_log - y_true_log) ** 2
+    return loss.mean(axis=0)
```
david-berthelot, line 71: For losses, we return the vector before the mean (so that it leaves freedom for manipulation by the user. For example to use thresholding like in FixMatch. It's also better for multi-GPU. See documentation Parallelism).

So in this case, you would simply do 
```
    return (y_pred_log - y_true_log) ** 2
```

Just for my education how would it work, say I have 2 vectors y_pred, y_true of shape (2, 3, 4, 5), what does the output loss look like?

AakashKumarNain:
Thanks for the review. A few points:
1. Can we [black](https://github.com/psf/black) for code-formatting? We use it heavily in `TF-addons`. In fact, the new keras website keras.io also uses `black`. It makes life much simpler
2. Will add the unit tests and updated the documentation. I wanted clarity on the code style.

objax/functional/loss.py
AakashKumarNain: I couldn't find this on wiki. But it is heavily used in `kaggle` competition as an evaluation metric. Here are the counterparts in `sklearn` and `TF` 
1. https://scikit-learn.org/stable/modules/model_evaluation.html#mean-squared-log-error
2. https://www.tensorflow.org/api_docs/python/tf/keras/losses/MSLE

objax/functional/loss.py
AakashKumarNain: The `mean` is always on the last axis. 
```
Input shape: (batch_size, d0, d1...dn)
Output shape: (batch_size, d0, d1...dn-1)
```

AakashKumarNain: I also noticed that I need not to put extra `1` as `log1p` takes cares of that. I will update the code. One question though: Even if we are doing `parallelism`, shouldn't we return the mean?

AakashKumarNain:
I formatted the code using `black` just to showcase ow it works. Lemme what you think about it. Thank you again for the reveiw

objax/functional/loss.py
david-berthelot: Thanks for the information. Well I guess there's no need to add anything then.

objax/functional/loss.py
```diff
@@ -48,6 +56,23 @@ def cross_entropy_logits_sparse(logits: JaxArray, labels: Union[JaxArray, int])
     return logsumexp(logits, axis=1) - logits[jn.arange(logits.shape[0]), labels]
 
 
+def mean_squared_logarithmic_error(y_true: JaxArray, y_pred: JaxArray) -> JaxArray:
+    """Computes the the mean squared logarithmic error.
+
```
david-berthelot, line 61: Here's a code sample of what I had in mind
```
def mean_squared_error(y_true: JaxArray, y_pred: JaxArray, keep_dims: Iterable[int] = (0,)) -> JaxArray:
    """Computes the mean squared logarithmic error between y_true and y_pred.
    
    Args:
        x: a tensor of shape (d0, .. dN-1).
        y: a tensor of shape (d0, .. dN-1).
        keep_dims: a sequence of the dimensions to keep.
        
    Returns:
        (d_i, ..., for i in keep_dims) tensor of the mean squared error.
    """
    loss = (jn.log1p(y_true) - jn.log1p(y_pred)) ** 2
    axis = [i for i in range(loss.ndim) if i not in keep_dims]
    return loss.mean(axis)
```

Using the default, it'll reduce on all axis except for 0 (which typically is the batch), which is standard in Objax. And the user can set keep_dims to their own taste if they want a different reduction.

AakashKumarNain:
Thank you for looking into it and the great suggestions. I have updated all the files.

I like the idea of providing the flexibility for `dims`. My only concern is that the semantics of `keep_dims` for an end-user may become confusing. For example, when we say `keepdims=True` in `numpy`, we expect that all dimensions would be preserved. And when we do an operation like `mean` or `sum`, instead of asking which dimensions to preserve, we pass the dimension to be squashed. If the vectors have many dimensions, it will become an easy point of doing mistake because passing the `axis` value is much more intuitive and along with the lines of other frameworks. 

One more thing that I want to point out. Is it really necessary to have `delta=1e-12`? Shouldn't it check for values upto `1e-7` or `1e-8` precision? I am asking this because if the numbers are very small, some tests can become tricky to verify upto that precision

david-berthelot:
> Thank you for looking into it and the great suggestions. I have updated all the files.

One more thing, I've created a PR for code formatting, you can look ahead on it (I tried pycharm formatting from the command line and it turns out it's working well and it's simple): https://github.com/google/objax/pull/22

> I like the idea of providing the flexibility for `dims`. My only concern is that the semantics of `keep_dims` for an end-user may become confusing. For example, when we say `keepdims=True` in `numpy`, we expect that all dimensions would be preserved. And when we do an operation like `mean` or `sum`, instead of asking which dimensions to preserve, we pass the dimension to be squashed. If the vectors have many dimensions, it will become an easy point of doing mistake because passing the `axis` value is much more intuitive and along with the lines of other frameworks.

Good points, I was wondering about the same thing. Passing the dimensions to squash was what I would have preferred but the problem is the input can have any number of dimensions, so there's no way to put a default that says "all dimensions except for dim 0". That's why I switched the API to say which dimensions to keep.

That's one of the difficulties, if some old/external decisions are limiting, then we should be ready to reform them. But, only when there's a benefit to it, otherwise there's no need to reinvent the wheel for the sake of it. This type of judgement is not easy to make.

Back to the question at hand. For a loss, sometimes we don't want to reduce on all dimensions (say for thresholding or even for weighting the entries in the batch or according to some dimension properties). Do you have any alternative in mind to `keep_dims`?

> One more thing that I want to point out. Is it really necessary to have `delta=1e-12`? Shouldn't it check for values upto `1e-7` or `1e-8` precision? I am asking this because if the numbers are very small, some tests can become tricky to verify upto that precision

Not it's not, the rule of thumb is to pick the *lowest delta that works* for the method/function under test.

So different functions will have different delta, for example when testing random.normal, we measure the variance of the generated random sample, it's very noisy and delta in that case is 1e-2 (or something like that, I don't recall exactly, but it's high).

Thanks

AakashKumarNain:
I have fixed the formatting as per #22 

> That's one of the difficulties, if some old/external decisions are limiting, then we should be ready to reform them. But, only when there's a benefit to it, otherwise there's no need to reinvent the wheel for the sake of it. This type of judgement is not easy to make.

Totally! These are some of the hardest decisions. 

> Do you have any alternative in mind to keep_dims?

I would say that instead of using `keep_dims`, we should simply use `dims`. Doing this:
1. Won't break the semantic meaning of `keep_dim` for the users as many libraries doing that. So, less confusion for the end user.
2. Because this is something new, we need to specify the semantics with examples in the docs

david-berthelot:
> I would say that instead of using `keep_dims`, we should simply use `dims`. Doing this:
> 
> 1. Won't break the semantic meaning of `keep_dim` for the users as many libraries doing that. So, less confusion for the end user.
> 2. Because this is something new, we need to specify the semantics with examples in the docs

So would `dims` indicate the dimensions that are left after the reduce or the dimensions to reduce on?
Here's another idea: `keep_axis`

We can also wait for Monday to have some more opinion from the Objax team.

AakashKumarNain:
> So would dims indicate the dimensions that are left after the reduce or the dimensions to reduce on?
That should be the team's decision purely. I like `keep_axis` much better than `dims`.

Sure. No rush. It's an interesting and important discussion.

Thank you

kihyuks:
> > So would dims indicate the dimensions that are left after the reduce or the dimensions to reduce on?
> 
> That should be the team's decision purely. I like `keep_axis` much better than `dims`.
> 
> Sure. No rush. It's an interesting and important discussion.
> 
> Thank you

keep_axis also looks good to me and I agree that we should provide a doc example for readers/users not to be confused its usage from keepdims or keep_dims.

david-berthelot:
Let's move ahead with `keep_axis` then.

parmarsuraj99:
Should we change the argument for other functions too (mae, mse)?

david-berthelot:
> Should we change the argument from `keep_dims` to `keep_axis` for other functions too (mae, mse)?

I've made PR for it already https://github.com/google/objax/pull/25

AakashKumarNain:
Thanks @david-berthelot Now that #25 has been merged, I will update the code

david-berthelot:
> Thanks @david-berthelot Now that #25 has been merged, I will update the code

Do you need help to wrap up this code submission? Let me know if you are too busy.

AakashKumarNain:
> > Thanks @david-berthelot Now that #25 has been merged, I will update the code
> 
> Do you need help to wrap up this code submission? Let me know if you are too busy.

I didn't get you. I have already made the final changes. Is there anything that you want me to change?

david-berthelot:
> > > Thanks @david-berthelot Now that #25 has been merged, I will update the code
> > 
> > 
> > Do you need help to wrap up this code submission? Let me know if you are too busy.
> 
> I didn't get you. I have already made the final changes. Is there anything that you want me to change?

No nothing else. I didn't realize you were waiting on the other branch to be merged. Thanks.

AakashKumarNain:
Haha no worries. Thank you for that fix (`keep_axis`) as well

david-berthelot (approved):

PR #19, opened 2020-08-29
Fix typos in Sphinx documentation pages

seungjaeryanlee:
This looks like an awesome framework! I was following the documentation and noticed some typos that I fixed in this PR.

### DIFF
--- a/docs/source/advanced/io.rst
+++ b/docs/source/advanced/io.rst
@@ -48,7 +48,7 @@ descriptor, so another way to save would be::
 
 .. note::
     The advantage of using a filename instead of file handle is that data will be written to a temporary file
-    first and temporary file will be renamed to provided filename only after all data has been written.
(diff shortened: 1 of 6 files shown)


docs/source/advanced/io.rst
```diff
@@ -48,7 +48,7 @@ descriptor, so another way to save would be::
 
 .. note::
     The advantage of using a filename instead of file handle is that data will be written to a temporary file
-    first and temporary file will be renamed to provided filename only after all data has been written.
+    first and a temporary file will be renamed to provided filename only after all data has been written.
```
david-berthelot, line 51: Maybe `and the temporary file` since it's the same file?

seungjaeryanlee: Good point! That definitely sounds smoother. Fixed it :)

PR #20, opened 2020-08-29
Add mean squared logarithmic loss function

parmarsuraj99:
Added mean_squared_error loss function

### DIFF
--- a/objax/functional/loss.py
+++ b/objax/functional/loss.py
@@ -80,6 +80,7 @@ def mean_absolute_error(x: JaxArray, y: JaxArray, keep_dims: Iterable[int] = (0,
     axis = [i for i in range(loss.ndim) if i not in keep_dims]
     return loss.mean(axis)
 
+
 def mean_squared_error(x: JaxArray, y: JaxArray, keep_dims: Iterable[int] = (0,)) -> JaxArray:
     """Computes the mean squared error between x and y.
     
--- a/objax/functional/loss.py
+++ b/objax/functional/loss.py
@@ -15,11 +15,11 @@
 __all__ = ['cross_entropy_logits', 
            'cross_entropy_logits_sparse', 
            'l2', 
-           'sigmoid_cross_entropy_logits', 
+           'mean_absolute_error',
            'mean_squared_error', 
-           'mean_absolute_error']
+           'sigmoid_cross_entropy_logits']
 
-from typing import Union
+from typing import Union, Iterable
 
 import jax.numpy as jn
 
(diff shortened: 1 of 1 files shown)


parmarsuraj99:
@googlebot I signed it!

objax/functional/loss.py
```diff
@@ -71,3 +71,16 @@ def sigmoid_cross_entropy_logits(logits: JaxArray, labels: Union[JaxArray, int])
         (batch, ...) tensor of the cross-entropies for each entry.
     """
     return jn.maximum(logits, 0) - logits * labels + jn.log(1 + jn.exp(-jn.abs(logits)))
+
+
+def mean_squared_error(logits: JaxArray, labels: JaxArray) -> JaxArray:
```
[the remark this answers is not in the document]
david-berthelot: Overall I would propose to remove the semantic assumptions:
- the loss works for general vectors, not just logits and labels, same comment about batch, also works if there's no batch.
- for more flexibility, the reduction should be customizable. Typically we don't reduce on axis 0 by default

Here's some sample code, let me know what you think about it:
```
def mean_squared_error(x: JaxArray, y: JaxArray, keep_dims: Iterable[int] = (0,)) -> JaxArray:
    """Computes the mean squared error between x and y.
    Args:
        x: a tensor of shape (d0, .. dN-1).
        y: a tensor of shape (d0, .. dN-1).
        keep_dims: a sequence of the dimensions to keep.
    Returns:
        (d_i, ..., for i in keep_dims) tensor of the mean absolute error.
    """
    loss = (x - y) ** 2
    axis = [i for i in range(loss.ndim) if i not in keep_dims]
    return loss.mean(axis)
```

And then trying it, you'd get:
```
>>> x = objax.random.normal((2, 3, 4, 5))
>>> y = objax.random.normal((2, 3, 4, 5))
>>> mean_squared_error(x, y, keep_dims=()).shape
()
>>> mean_squared_error(x, y).shape
(2,)
>>> mean_squared_error(x, y, keep_dims=(0, 1)).shape
(2, 3)
```

objax/functional/loss.py
```diff
@@ -71,3 +76,29 @@ def sigmoid_cross_entropy_logits(logits: JaxArray, labels: Union[JaxArray, int])
...
+        logits: (batch_size, d0, .. dN) tensor of the predicted values.
+        labels: (batch_size, d0, .. dN) tensor of the ground truth values.
+
+    Returns:
+        (batch_size, d0, .. dN-1) tensor of mean squared error for each entry.
+    """
+    return jn.mean(jn.square(labels - logits), axis=-1)
+
+
+def mean_absolute_error(logits: JaxArray, labels: JaxArray) -> JaxArray:
```
david-berthelot, line 94: Same general comments as for mean squared error.

parmarsuraj99 (commented):
Formatted

objax/functional/loss.py
parmarsuraj99: Thanks. Updated both functions as suggested.

PR #22, opened 2020-08-30
Style rules for PyCharm and style discussion in CONTRIBUTING.md.

david-berthelot:


AakashKumarNain:
This looks good to me. Though there is one concern. Many people, including myself, don't use `pycharm` (Ik, ik it sounds stupid but I am more of `vim`, `notebook` and `vscode` guy). Anything that is editor specific would do harm. 

For example, at some point in time (and I am very positive that it is going to be very soon), `objax` would have a very high traction. There would be many many PRs and we would require a `CI/CD` for managing them. Forcing this `code-style` check configured to a specific editor wouldn't be easy. 

So, if we don't want to use `black`, then it's fine but we can use `flake8` for keeping the consistency for all the end users. This is just my opinion

david-berthelot:
> This looks good to me. Though there is one concern. Many people, including myself, don't use `pycharm` (Ik, ik it sounds stupid but I am more of `vim`, `notebook` and `vscode` guy). Anything that is editor specific would do harm.
> 
> For example, at some point in time (and I am very positive that it is going to be very soon), `objax` would have a very high traction. There would be many many PRs and we would require a `CI/CD` for managing them. Forcing this `code-style` check configured to a specific editor wouldn't be easy.
> 
> So, it's fine if we don't want to use `black`, but we can use `flake8` for keeping the consistency for all the end users. This is just my opinion

I totally agree with your comments, I'm not super familiar with python formatting systems. But if there's a standalone formatter that produces an output like PyCharm formatting (which is concise and pleasing to the eye), then I'm all for it.

As to flake8, that would be helpful (I get it's a formatting checking tool). Do you mind filing a feature request for it?

AakashKumarNain:
`flake8` is super simple and similar to what you are looking for. You can read about it in detail [here](https://flake8.pycqa.org/en/latest/). I will file a feature request for it but wouldn't be able to do it today (already very late in my timzone)

.editorconfig
```diff
@@ -0,0 +1,79 @@
+[*]
```
AlexeyKurakin, line 1: I assume this file is PyCharm config.
Could you add comment in the beginning of the file, explaining what it is?

david-berthelot: Actually no, it's a general config format. That said it has some special settings for pycharm too.

david-berthelot: Done

PR #27, opened 2020-09-05
Bugfix in average_pool2d, missing divide by the sum of strides.

david-berthelot:


tests/functional_pooling.py
```diff
@@ -27,6 +27,18 @@ def shaparange(s):
 
 
 class TestPooling(unittest.TestCase):
+    def test_average_pooling2d(self):
+        """Test batch_to_space2d and space_to_batch2d."""
```
kihyuks, line 31: Change docstring into Test average pooling 2d

tests/functional_pooling.py
```diff
@@ -27,6 +27,18 @@ def shaparange(s):
...
+        x = shaparange((2, 3, 10, 30))
+        y = objax.functional.average_pool_2d(x, size=5)
+        z = x.reshape((2, 3, 2, 5, 6, 5)).mean((-3, -1))
+        self.assertEqual(y.tolist(), z.tolist())
+        y = objax.functional.average_pool_2d(x, size=5, strides=1)
```
kihyuks, line 36: Can we have a test with a small tensor to check the output value of average_pool_2d analytically when the size and strides are different?

kihyuks (changes requested):

tests/functional_pooling.py
david-berthelot: Done

tests/functional_pooling.py
david-berthelot: Actually I removed the docstring, it's redundant with the test name and provided no useful information.

tests/functional_pooling.py
david-berthelot: Done

PR #39, opened 2020-09-14
Addition to objax.util.image:

david-berthelot:
- Extended to_png support to JaxArray.
- Added from_file API to read an image from a file.
- Added image_grid API to export a grid of images to an image.


objax/util/image.py
```diff
@@ -12,17 +12,37 @@
...
+
+    Return:
+        3D numpy array (C, H, W) normalized with normalize_to_unit_float.
+    """
+    image = np.asarray(Image.open(file).convert('RGB'))
+    return normalize_to_unit_float(image.transpose((2, 0, 1)))
+
+
+def image_grid(image: np.ndarray) -> np.ndarray:
+    """Take an image grid (nh, nw, gw, c, h, w) and export it to an image (c, nh * h, nh * w)."""
```
AlexeyKurakin, line 41: What is `gw`? From the code of the function it looks like input image should have 5 ndims, not 6.

I assume that this the main purpose of this function for visualisation (take a batch (nh, nw) of images and convert it into image grid. In this case maybe better explanation would be something like following:

```Rearrange array of images (nh, hw, c, h, w) into image grid on single image.```

i.e. basically emphasize that input is batch/matrix/array of images and output is single image where source images are arranged as a grid.

AlexeyKurakin (approved):

objax/util/image.py
david-berthelot: Thanks, done.

Issue #40, opened 2020-09-14
Naming of the `GradValues` function


jli05:
If I understand right, `GradValues` essentially does two things: computing gradients and computing model final values.

So why not split it into two functions? Or if we keep the current form, could we name it `GradAndValuesFn`? Just thinking this is a prominent function and want to keep it the easiest for people beginning to use the framework. An easy name as `fit()` and `predict()` made `scikit-learn`.

AlexeyKurakin:
Yes, `GradValues` returns both gradient and value of the input function/module.

It's pretty much Objax wrapper for [jax.value_and_grad](https://jax.readthedocs.io/en/latest/jax.html#jax.value_and_grad) function. This is where naming comes from.

The reason why this module returns both gradient and value for following reasons:

1) To compute gradient, forward pass has to be performed. Thus value of the function is available for free.
2) Quite often, user need both value and gradient. Example: when compute both value and gradient of loss function during training. In such case returning both gradient and value is a performance optimization, compared to computing gradient and value separately.

As an option we can add module `Grad` which will only returns gradient.
`Grad` could be implemented either as a wrapper for [jax.grad](https://jax.readthedocs.io/en/latest/jax.html#jax.grad) or by calling `GradValues` and throwing away value of the function.
In both cases (wrapper of `jax.grad` or using `GradValues` internally), `Grad` will be computing both value and gradient and throwing away value of the function.

Let us know if module `Grad` will be useful and make things more clear.
Also if you willing to contribute, feel free to implement it and submit a pull request.
Implementation of [GradValues](https://github.com/google/objax/blob/ddcdd30e00575dd2bfcffa988ea7b4388bd1ff8a/objax/gradient.py#L26) could be used as a reference.

sathish-a:
@jli05 thanks for bringing up this. @AlexeyKurakin I've implemented the `Grad` by calling `GradValues` internally and returned only the gradients. I tested it with MNIST code in the examples just to check If I'm affecting any other flows and it worked fine. Now that I couldn't find any guidelines about what else to be added like (creating .rst, etc.) if a new module is introduced. Can you help me with it?

jli05:
Why the forward function valuation is "needed" for computing the gradients whilst mathematically it doesn't necessarily look so?

I wrote the question:

1. Engineering-wise, it seems better to split a function which does two things into two that does one each, so that we give choice, esp. the value part can already be given by `model(input)` thus may seem "redundant".

[...]

PR #43, opened 2020-09-17
Add module Grad

sathish-a:
With reference to #40. Added a module `Grad`. `Grad` internally calls `GradValues` and return only the gradients.
* No linter error.
* All tests passed.

### DIFF
--- a/objax/gradient.py
+++ b/objax/gradient.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__all__ = ['GradValues']
+__all__ = ['GradValues', 'Grad']
 
 from typing import List, Optional, Callable, Tuple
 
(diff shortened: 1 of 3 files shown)


objax/gradient.py
```diff
@@ -71,3 +71,28 @@ def __call__(self, *args):
                                        list(args))
         self.vc.subset(BaseState).assign(changes)
         return g, outputs
+
+
+class Grad(ModuleWrapper):
```
david-berthelot, line 76: Can you move the new function below GradValues (we try to keep the order alphabetical within files)?

objax/gradient.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__all__ = ['GradValues']
+__all__ = ['GradValues', 'Grad']
```
david-berthelot, line 15: Suggestion `Grad` followed by `GradValues` (see next comment).

david-berthelot (changes requested):
Can you also and entry to https://github.com/google/objax/blob/master/docs/source/objax/objax.rst
for the API to show in docs?

Thanks!

objax/gradient.py
sathish-a: Sure. This applies to the test script also right?

david-berthelot (approved):

PR #45, opened 2020-09-18
Add proper support for multi-host distributed setup

AlexeyKurakin:


objax/io/checkpoint.py
```diff
@@ -56,11 +62,14 @@ def __init__(self, logdir: str, keep_ckpts: int, makedir: bool = True, verbose:
             makedir: if True then directory for checkpoints will be created,
                 otherwise it's expected that directory already exists.
             verbose: if True then print when data is restored from checkpoint.
+            only_first_host_writes: if True then checkpoints are saved only by the first host in multi-host setup,
```
david-berthelot, line 65: Do I understand correctly: the code does not work if the option is set to `False`?
If that's indeed correct, why even make this an option?

objax/io/checkpoint.py
```diff
@@ -107,6 +116,8 @@ def save(self, vc: VarCollection, idx: int):
             vc: variables collection to save.
             idx: index of the new checkpoint where variables should be saved.
         """
+        if not self.do_writes:
```
david-berthelot, line 119: On second thought, I'm wondering if modifying `checkpoint.py` is the right thing to do. It adds complexity and it's not intuitive. In fact, it seems better suited to only save in imagenet training loop if `host_id == 0`.

objax/jaxboard.py
```diff
@@ -95,27 +96,36 @@ def __call__(self):
 class SummaryWriter:
     """Writes entries to event files in the logdir to be consumed by Tensorboard."""
 
-    def __init__(self, logdir: str, queue_size: int = 5, write_interval: int = 5):
+    def __init__(self, logdir: str, queue_size: int = 5, write_interval: int = 5, only_first_host_writes: bool = True):
```
david-berthelot, line 99: Again, this multi-host setup trickles down every module and pollutes them. I think it should be handled at a higher level if possible.

david-berthelot (changes requested):

objax/jaxboard.py
AlexeyKurakin: I think there are few options how to handle multi-host setup:

A. check `jax.host_id() == 0` in the training loop. This won't pollute `jaxboard.py` and `checkpoint.py`, however this would require user to always modify training loop

B. checkpoint and jaxboard perform writes only when `jax.host_id() == 0`.

C. checkpoint and jaxboard BY DEFAULT perform writes only when `jax.host_id() == 0`. However this behavior could be changed/altered. <-- this was original code of the pull request


Between options B and C, option B actually covers 99.99999% of cases, but theoretically I could imagine some weird situation when user may want to manually control which host perform writes (some weird model parallelism setup where several hosts actually perform writes, but they write different things).

Between option A and BC:
* **Advantage of options B and C**: any single-host code will work on multi-host setup pretty much without modifications. On the other hand, option A would require users to manually add `jax.host_id() == 0` on each IO op.
    * Theoretically it's better to add parallel barrier to the end of the multi-host program. But in practice code will work without parallel barrier, just last training step may be corrupted.
* **Advantage of option A**: it keeps the code of jaxboard and checkpoint cleaner.


WDYT?

david-berthelot: In light of your comments I prefer A.

AlexeyKurakin:
Sounds good
I'll create a new pull request with updated change, since I messed up state of my local git

PR #52, opened 2020-09-22
Raise nicer error if parallel can't divide batch by num devices

carlini:
If you try and run a parallel module over a batch of examples where the batch shape doesn't divide the number of devices, the code crashes in numpy land. Instead, throw a nicer error message.

### DIFF
--- a/objax/module.py
+++ b/objax/module.py
@@ -191,6 +191,8 @@ class Parallel(ModuleWrapper):
 
     def device_reshape(self, x: JaxArray) -> JaxArray:
         """Utility to reshape an input array in order to broadcast to multiple devices."""
+        assert x.shape[0]%self.ndevices == 0, f'Must be able to equally divide batch {x.shape} among ' \
+                                              f'{self.ndevices} devices, but does not go equally.'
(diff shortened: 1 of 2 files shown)


objax/module.py
```diff
@@ -191,6 +191,8 @@ def pmap(tensor_list: List[ShardedDeviceArray], random_list: List[ShardedDeviceA
 
     def device_reshape(self, x: JaxArray) -> JaxArray:
         """Utility to reshape an input array in order to broadcast to multiple devices."""
+        assert x.shape[0]%self.ndevices == 0, f'Must be able to equally divide batch {x.shape} among ' \
```
david-berthelot, line 194: Nit: spacing, can you put space around operators to be consistent with rest of the code?
```python
assert x.shape[0] % self.ndevices == 0, f'Must be able to equally divide batch {x.shape} among ' \
                                        f'{self.ndevices} devices, but does not go equally.'
```

david-berthelot (changes requested):

david-berthelot (approved):

PR #54, opened 2020-09-23
Naming manipulation tools to facilitate importing model weights from …

### LINKED ISSUE #42: More control over var/module namespace.
I got my first 'hello world' model experiment working w/ Objax. I adapted my PyTorch EfficientNet impl. Overall pretty smooth, currently wrapping Conv2d so I can get the padding I want.

One thing that stuck out after inspecting the model, the var namespace is a mess. An aspect of modelling that I value highly is the ability to have sensible checkpoint/var maps to work with. I often end up dealing with conversions between frameworks, exports for mobile or embedded targets and having your vars (parameters) sensibly named, and often being able to control those names in the originating framework is important. 

Any thoughts on improving this? The current name/scoping mechanism forces the inclusion of the Module class names, is that necessary? Shouldn't attr names through the tree be enough for uniqueness?

Also, there is no ability to specify names for modules in sequential containers. I use this quite often for frameworks that have it. Sometimes I don't care much (long list of block repeats, 0..n is fine), but for finer grained blocks I like to know what conv is what by looking at the var names. '0.b, o.w' etc isn't very useful.

I'll post an example of the var keys below, and comparison point for pytorch.

david-berthelot:
…other frameworks and to remap weights between models.

- VarCollection has a new rename method to rename its variables.
- load_var_collection has new option to rename variables from the data read from the file.
- Renamer allows to take a dictionary of strings {source:target}, or a tuple (regex, replacement) or a function (str -> str).

Resolves #42


objax/util/util.py
```diff
@@ -28,11 +30,30 @@
...
     """Custom dictionary that allows to access dict values as attributes."""
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.__dict__ = self
 
 
+class Renamer:
+    def __init__(self,
+                 rules: Union[Dict[str, str], Tuple[Pattern[str], str], Callable[[str], str]],
```
AlexeyKurakin, line 41: I think second option for `rules` type should be `List[Tuple[Pattern[str], str]]` or `Sequence[Tuple[Pattern[str], str]]` instead of `Tuple[Pattern[str], str]`

`Tuple[Pattern[str], str]` allows user to perform substitutions which match only one pattern, while I would expect that it's likely that user may have multiple patterns for substitution.

Also, `Dict[str, str]` allows user to pass dictionary with multiple substitutions, `Callable` allows to pass a function which can do arbitrary things. Compared to them `Tuple[Pattern[str], str]` looks like somewhat limiting.

AlexeyKurakin:
Overall design looks good to me. I had only one comment regarding arguments of `Renamer`

objax/util/util.py
david-berthelot: I agree `Sequence` is a lot nicer, and will do it in the final implementation if design is satisfactory.

tests/util.py
```diff
@@ -31,6 +31,34 @@ def test_easy_dict(self):
...
+
+        r = objax.util.Renamer([(re.compile('a{2}'), 'x'), (re.compile('bb'), 'y')])
+        self.assertEqual(r('baab'), 'bxb')
+        self.assertEqual(r('baaab'), 'bxab')
+        self.assertEqual(r('baaaab'), 'bxxb')
+        self.assertEqual(r('abba'), 'aya')
+        self.assertEqual(r('acca'), 'acca')
+
+        r = objax.util.Renamer([(re.compile('a{2}'), 'x'), (re.compile('xa'), 'y')])
+        self.assertEqual(r('baaab'), 'byb')
```
AlexeyKurakin, line 60: Add tests for chained renamer

objax/util/util.py
```diff
@@ -34,6 +36,38 @@ def __init__(self, *args, **kwargs):
...
+        Args:
+            rules: the replacement mapping.
+            chain: optionally, another renamer to call after this one completes.
+        """
+        self.chain = chain
+        if callable(rules):
+            self.subfn = rules
+        elif isinstance(rules, dict):
+            regex = re.compile('(%s)' % '|'.join(map(re.escape, rules.keys())))
+            self.subfn = functools.partial(regex.sub, lambda m: rules[m.string[m.start():m.end()]])
```
AlexeyKurakin, line 56: I think `m.string[m.start():m.end()]` could be simplified as `m.group(0)`
see https://docs.python.org/3/library/re.html#re.Match.group

AlexeyKurakin (commented):
Some minor comments in the code.

Also consider adding tests for `load_var_collection` with renamer

PR #55, opened 2020-09-23
Make objax.nn.Sequential more powerful.

david-berthelot:
- Support slicing to run part of the network.
- Support modules with arbitrary number of inputs/outputs.

### DIFF
--- a/docs/source/objax/nn.rst
+++ b/docs/source/objax/nn.rst
@@ -122,7 +122,7 @@ objax.nn
         ml = objax.nn.Sequential([objax.nn.Linear(2, 3), objax.functional.relu,
                                   objax.nn.Linear(3, 4)])
         x = objax.random.normal((10, 2))
-        y = ml(x)       # Runs all the operations (Linear -> ReLU -> BatchNorm0D -> Linear).
+        y = ml(x)       # Runs all the operations (Linear -> ReLU -> Linear).
         print(y.shape)  # (10, 4)
 
         # objax.nn.Sequential is really a list.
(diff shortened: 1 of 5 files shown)


docs/source/objax/nn.rst
```diff
@@ -122,12 +122,18 @@ objax.nn
         ml = objax.nn.Sequential([objax.nn.Linear(2, 3), objax.functional.relu,
                                   objax.nn.Linear(3, 4)])
         x = objax.random.normal((10, 2))
-        y = ml(x)  # Runs all the operations (Linear -> ReLU -> Linear).
+        y = ml(x)       # Runs all the operations (Linear -> ReLU -> BatchNorm0D -> Linear).
```
AlexeyKurakin, line 125: I don't see `BatchNorm0D` in the definition of `ml` two lines above

objax/module.py
```diff
@@ -74,6 +74,12 @@ def vars(self, scope: str = '') -> VarCollection:
                 vc.update(v.vars(scope=f'{scope}[{p}]'))
         return vc
 
+    def __getitem__(self, key: Union[int, slice]):
```
AlexeyKurakin, line 77: Do we need docstring for RST documentation?

david-berthelot: I don't think so. It's the standard list square bracket operator overloaded so it doesn't return a list but instead a ModuleList or Sequential. It's not meant to show up in the documentation either.

docs/source/objax/nn.rst
david-berthelot: Fixed

AlexeyKurakin (approved):

Issue #37, opened 2020-09-09
Easy support for per-module train/eval state


AlexeyKurakin:
This is tracking issue to improve support of per-module train/eval state in Objax.
This issue originates from discussion of PyTorch vs Objax-style of propagating training/eval mode https://github.com/google/objax/issues/29

PyTorch style allows easy way to specify per-module train/eval mode, like in the following example:
```python
model = Resnet50(nclasses=1000)
…
# Here is example how to set most of the network,
# except few modules into training mode
model.train()
model.block_1.bn_1.eval()
model.block_2.bn_2.eval()
```

There is no clean way to achieve the same thing in Objax right now. One possibility is to use `functools.partial`:
```python
model = Resnet50(nclasses=1000)
…
# Here is example how to force certain batch norms into eval mode
model.block_1.bn_1 = functools.partial(model.block_1.bn_1, training=False)
model.block_2.bn_2 = functools.partial(model.block_2.bn_2, training=False)

# following line will call model in training mode, except for two block_1.bn_1 and block_2.bn_2
y = model(x, training=True)
```

However there are some problems with `functools.partial`:
* it converts everything into a function (thus `vars()` are not propagated).
* if `bn_eval = functools.partial(bn, training=False)` and caller will try to pass `training` argument to `bn_eval` it will cause run-time error
* there is no easy way to undo `functools.partial` after it applied to a module

Thus we need a better solution to do per-module train/eval state

AlexeyKurakin:
Here is one idea of how this could be implemented:

```
######### in module.py

class ArgsOverride(Module):

    def __init__(self, base_module, **kwargs):
        self.base_module = base_module
        self.kwargs = kwargs
  
    def vars(self, scope):
        # presence of ArgsOverride module won’t affect variable names
        return self.base_module.vars(scope)

    def __call__(self, *args, **kwargs):
        kwargs.update(self.kwargs)  # in practice this should only
                                    # kwargs which are presented in
                                    # signature of base_module
        self.base_module(*args, **kwargs)

######### in utils.py

def reset_args_override(module):
  # removes ArgsOverride from module and all submodules
  … 

############ in user code

model = Resnet50(nclasses=1000)
…
# Here is example how to set most of the network,
# except few modules into training mode
model.block_1.bn_1 = objax.ArgsOverride(model.block_1.bn_1, training=False)
model.block_2.bn_2 = objax.ArgsOverride(model.block_2.bn_2, training=False)

# reset all args override
[...]

PR #73, opened 2020-09-25
Adding ForceArgs module

AlexeyKurakin:
This address issue #37

### DIFF
--- a/objax/module.py
+++ b/objax/module.py
@@ -22,7 +22,7 @@ import jax.numpy as jn
 from jax.interpreters.pxla import ShardedDeviceArray
 
 from objax.typing import JaxArray
-from objax.util import merge_args_kwargs, positional_args_names
+from objax.util import override_args_kwargs, positional_args_names
 from objax.variable import BaseState, BaseVar, RandomState, VarCollection
 
 
@@ -69,36 +69,35 @@ class ForceArgs(Module):
             for idx, v in enumerate(module):
                 if isinstance(v, Module):
                     if isinstance(v, ForceArgs):
-                        v = v._wrapped
+                        v = v.__wrapped__
                         module[idx] = v
                     ForceArgs.undo(v)
         else:
             for k, v in module.__dict__.items():
                 if isinstance(v, Module):
                     if isinstance(v, ForceArgs):
-                        v = v._wrapped
-                        module.__dict__[k] = v
+                        v = v.__wrapped__
+                        setattr(module, k, v)
                     ForceArgs.undo(v)
 
     def __init__(self, module: Module, **kwargs):
         """Initializes ForceArgs.
 
         Args:
-            module: base module, which argument will be overridden.
+            module: module which argument will be overridden.
             kwargs: values of keyword arguments which will be forced to use.
         """
         if isinstance(module, ForceArgs):
             raise ValueError('Can not force arguments on ForceArgs module.')
-        self._wrapped = module
+        self.__wrapped__ = module
         self.forced_kwargs = kwargs
 
     def vars(self, scope: str = '') -> VarCollection:
-        return self._wrapped.vars(scope=scope)
+        return self.__wrapped__.vars(scope=scope)
 
     def __call__(self, *args, **kwargs):
-        kwargs.update(self.forced_kwargs)
-        args, kwargs = merge_args_kwargs(args, kwargs, self._wrapped)
-        return self._wrapped(*args, **kwargs)
+        args, kwargs = override_args_kwargs(self.__wrapped__, args, kwargs, self.forced_kwargs)
+        return self.__wrapped__(*args, **kwargs)
 
 
 class ModuleList(Module, list):
--- a/objax/module.py
+++ b/objax/module.py
@@ -15,7 +15,7 @@
 __all__ = ['ForceArgs', 'Function', 'Jit', 'Module', 'ModuleList', 'Parallel', 'Vectorize']
 
 from collections import namedtuple
-from typing import Any, Dict, Optional, List, Union, Callable, Tuple
+from typing import Optional, List, Union, Callable, Tuple
 
 import jax
 import jax.numpy as jn
@@ -59,14 +59,6 @@ class ForceArgs(Module):
     ANY = namedtuple('Any', ())
     """Token used in `ForceArgs.undo` to indicate undo of all values of specific argument."""
 
-    def _remove_args(self, args_to_remove: Dict[str, Any]):
-        """Removes given arguments from override list of `ForceArgs` instance."""
-        if not args_to_remove:
-            self.forced_kwargs = {}
-        else:
-            self.forced_kwargs = {k: v for k, v in self.forced_kwargs.items()
-                                  if (k not in args_to_remove) or (args_to_remove[k] not in (v, ForceArgs.ANY))}
-
     @staticmethod
     def undo(module: Module, **kwargs):
         """Undo ForceArgs on each submodule of the module. Modifications are done in-place.
@@ -79,7 +71,11 @@ class ForceArgs(Module):
                 If `**kwargs` is empty then all overrides will be undone.
         """
         if isinstance(module, ForceArgs):
-            module._remove_args(kwargs)
+            if not kwargs:
+                module.forced_kwargs = {}
+            else:
+                module.forced_kwargs = {k: v for k, v in module.forced_kwargs.items()
+                                        if (k not in kwargs) or (kwargs[k] not in (v, ForceArgs.ANY))}
             ForceArgs.undo(module.__wrapped__, **kwargs)
         elif isinstance(module, ModuleList):
             for idx, v in enumerate(module):
--- a/objax/module.py
+++ b/objax/module.py
@@ -14,6 +14,7 @@
 
 __all__ = ['ForceArgs', 'Jit', 'Module', 'ModuleList', 'ModuleWrapper', 'Parallel', 'Vectorize']
 
+from collections import namedtuple
 from types import MethodType
 from typing import Optional, List, Union, Callable, Tuple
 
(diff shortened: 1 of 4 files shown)


objax/module.py
```diff
@@ -53,6 +53,54 @@ def __call__(self, *args, **kwargs):
...
+                        v = v._wrapped
+                        module[idx] = v
+                    ForceArgs.undo(v)
+        else:
+            for k, v in module.__dict__.items():
+                if isinstance(v, Module):
+                    if isinstance(v, ForceArgs):
+                        v = v._wrapped
+                        module.__dict__[k] = v
+                    ForceArgs.undo(v)
```
david-berthelot, line 81: The code can be greatly simplified noting that:
1. `ForceArgs` called on a root of module tree, returns a new module.
2. Consequently calling undo on this root should also return a module, the original one.

Also I think `setattr` is the preferred way to write attributes in Python, so I also made that change.

```suggestion
    def undo(module: Module):
        """Undo ForceArgs on each submodule of the module.

        Args:
            module: module for which to undo ForceArgs.
        """
        if isinstance(module, ForceArgs):
            return ForceArgs.undo(module.__wrapped__)

        if isinstance(module, ModuleList):
            for idx, v in enumerate(module):
                if isinstance(v, Module):
                    module[idx] = ForceArgs.undo(v)
            return module

        for k, v in module.__dict__.items():
            if isinstance(v, Module):
                setattr(module, k, ForceArgs.undo(v))
        return module
```

objax/module.py
```diff
@@ -53,6 +53,54 @@ def __call__(self, *args, **kwargs):
...
+        """Initializes ForceArgs.
+
+        Args:
+            module: base module, which argument will be overridden.
+            kwargs: values of keyword arguments which will be forced to use.
+        """
+        if isinstance(module, ForceArgs):
+            raise ValueError('Can not force arguments on ForceArgs module.')
+        self._wrapped = module
+        self.forced_kwargs = kwargs
```
david-berthelot, line 93: Two things:
1. `base_module` can be simply called `module`.
2. Rename `_wrapped` to `__wrapped__` which is used by python to determine the signature of a function.
```suggestion
    def __init__(self, module: Module, **kwargs):
        """Initializes ForceArgs.

        Args:
            module: module which arguments will be overridden.
            kwargs: values of keyword arguments which will be forced to use.
        """
        if isinstance(module, ForceArgs):
            raise ValueError('Can not force arguments on ForceArgs module.')
        self.__wrapped__ = module
        self.forced_kwargs = kwargs
```

objax/module.py
```diff
@@ -53,6 +53,54 @@ def __call__(self, *args, **kwargs):
...
+
+    def vars(self, scope: str = '') -> VarCollection:
+        return self._wrapped.vars(scope=scope)
```
david-berthelot, line 96:
```suggestion
        return self.__wrapped__.vars(scope=scope)
```

objax/module.py
```diff
@@ -53,6 +53,54 @@ def __call__(self, *args, **kwargs):
...
+
+    def __call__(self, *args, **kwargs):
+        kwargs.update(self.forced_kwargs)
+        args, kwargs = merge_args_kwargs(args, kwargs, self._wrapped)
+        return self._wrapped(*args, **kwargs)
```
david-berthelot, line 101: The current method had some issue with Python, I explain it in details on the suggestion for `override_args_kwargs`.
```suggestion
        args, kwargs = override_args_kwargs(args, kwargs, self.forced_kwargs, self.__wrapped__)
        return self.__wrapped__(*args, **kwargs)
```

objax/module.py
```diff
@@ -22,7 +22,7 @@
 from jax.interpreters.pxla import ShardedDeviceArray
 
 from objax.typing import JaxArray
-from objax.util import positional_args_names
+from objax.util import merge_args_kwargs, positional_args_names
```
david-berthelot, line 25: See code comments below to see why I changed the function (it's not just a rename).
```suggestion
from objax.util import overridee_args_kwargs, positional_args_names
```

objax/util/util.py
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 __all__ = ['EasyDict', 'args_indexes', 'dummy_context_mgr', 'ilog2', 'local_kwargs', 'map_to_device',
-           'multi_host_barrier', 'positional_args_names', 'to_tuple']
+           'merge_args_kwargs', 'multi_host_barrier', 'positional_args_names', 'to_tuple']
```
david-berthelot, line 16:
```suggestion
           'override_args_kwargs', 'multi_host_barrier', 'positional_args_names', 'to_tuple']
```

objax/util/util.py
```diff
@@ -69,6 +69,32 @@ def local_kwargs(kwargs: dict, f: Callable) -> dict:
...
+    """
+    args = list(args)
+    kwargs = local_kwargs(kwargs, f)
+    p = inspect.signature(f).parameters
+    for idx, (k, v) in enumerate(p.items()):
+        if (idx >= len(args)) or (v.kind == inspect.Parameter.VAR_POSITIONAL):
+            break
+        if k in kwargs:
+            args[idx] = kwargs.pop(k)
+    return args, kwargs
```
david-berthelot, line 95: I propose a significant change here since the original implementation was hiding code bugs.

Here's an example:
```python
class M(Module):
    def __call__(self, a, b, c):
        return 0

g = objax.ForceArgs(M(), b=1)
g(1, 2, 3, b=4) # should error but didn't in the original code.
f(1, 2, 3, b=4) # the ForceArgs version should error just like the original function f.
```

There were other failure scenarios. But the gist is that when passing improper/invalid argument, the behavior of a `ForceArgs(f)` should be the one of `f`. Consequently, I made the following suggestion to only override the value of an argument if present without modifying arguments in any other way.

```suggestion
def override_args_kwargs(args: Tuple[Any, ...],
                         kwargs: Dict[str, Any],
                         override:  Dict[str, Any],
                         f: Callable) -> Tuple[List, Dict[str, Any]]:
    """Override positional and keyword arguments according to signature of the function.

    Replace the values in args and kwargs with the corresponding override entries.

    Args:
        args: positional arguments.
        kwargs: keyword arguments.
        override: arguments for which to override the value.
        f: callable, for which arguments should be fixed.

    Return:
        args: updated list of positional arguments.
        kwargs: updated dictionary of keyword arguments.
    """
    args = list(args)
    params = inspect.signature(f).parameters
    override = override.copy()
    for i, (k, v) in enumerate(itertools.islice(params.items(), len(args))):
        if v.kind not in (inspect.Parameter.POSITIONAL_ONLY, inspect.Parameter.POSITIONAL_OR_KEYWORD):
            break
        if k in override:
            args[i] = override.pop(k)
    return args, {**kwargs, **override}
```

tests/module.py
```diff
@@ -75,6 +107,24 @@ def test_module_wrapper(self):
...
+        np.testing.assert_almost_equal(model(x, some_arg=1.0), [6., 7.])
+        # set forced args in a list
+        model.block1.seq[0] = objax.ForceArgs(model.block1.seq[0], some_arg=-1.0)
+        np.testing.assert_almost_equal(model(x, some_arg=0.0), [-1., 0.])
+        np.testing.assert_almost_equal(model(x, some_arg=1.0), [4., 5.])
+        # undo force args
+        objax.ForceArgs.undo(model)
+        np.testing.assert_almost_equal(model(x, some_arg=0.0), [1., 2.])
+        np.testing.assert_almost_equal(model(x, some_arg=1.0), [8., 9.])
+
```
david-berthelot, line 127: Add tests for negative cases with `self.assertRaises` to make sure a `ForceArgs(f)` produces the same errors as `f` when called improperly (there was one example in a previous suggestion).

david-berthelot (changes requested):

objax/module.py
AlexeyKurakin: Done

objax/module.py
AlexeyKurakin: Done

objax/util/util.py
AlexeyKurakin: Done

objax/module.py
AlexeyKurakin: I think user should not be allowed to call `ForceArgs.undo` on instance of `ForceArgs` because it may lead to surprising behavior from user point of view:

```python
# override args of one operation in the model
m = SomeModel()
m.bn1 = ForceArgs(m.bn1, training=False)
y = m(x, training=True)  # argument of m.bn1 is overriden

ForceArgs.undo(m)  # reset all overrides

# override args of entire model
pred_m = ForceArgs(m, training=False)
y_pred = pred_m(x)  # training is overriden for entire model

# if ForceArgs.undo can be called on instance of ForceArgs following line will do nothing
# however user may expect that arguments override will be undone
ForceArgs.undo(pred_m)
```

There is a separate question whether `ForceArgs.undo(m)` should return `m`.
I think it's ok to add, but not sure how useful. I just allow users to write code like `y = ForceArgs.undo(m)(x)` which might be actually somewhat confusing

AlexeyKurakin:
@david-berthelot code updated, please take a look

objax/module.py
david-berthelot: 1. One can call `ForceArgs` on the root, therefore calling `ForceArgs.undo` should be doable on the root too.
2. Returning a module results in simpler implementation too.

david-berthelot (changes requested):
The `undo` method looks complicated and doesn't work on the root.

AlexeyKurakin:
@david-berthelot ForceArgs updated. Please take a look

objax/module.py
```diff
@@ -53,6 +54,68 @@ def __call__(self, *args, **kwargs):
...
 
 
+class ForceArgs(Module):
+    """Forces override of arguments of given module."""
+
+    ANY = namedtuple('Any', ())
+    """Token used in `ForceArgs.undo` to indicate undo of all values of specific argument."""
+
+    @staticmethod
+    def _remove_args(force_args, args_to_remove: dict):
```
david-berthelot, line 64:
```suggestion
    def _remove_args(force_args: 'ForceArgs', args_to_remove: Dict[str, Any]):
```

objax/module.py
```diff
@@ -53,6 +54,68 @@ def __call__(self, *args, **kwargs):
...
+        if isinstance(module, ForceArgs):
+            ForceArgs._remove_args(module, kwargs)
+            ForceArgs.undo(module.__wrapped__, **kwargs)
+        elif isinstance(module, ModuleList):
+            for idx, v in enumerate(module):
+                if isinstance(v, Module):
+                    ForceArgs.undo(v, **kwargs)
+                    if isinstance(v, ForceArgs) and not v.forced_kwargs:
+                        v = v.__wrapped__
+                        module[idx] = v
```
david-berthelot, line 92:
```suggestion
                        module[idx] = v.__wrapped__
```

objax/module.py
```diff
@@ -53,6 +54,68 @@ def __call__(self, *args, **kwargs):
...
+        else:
+            for k, v in module.__dict__.items():
+                if isinstance(v, Module):
+                    ForceArgs.undo(v, **kwargs)
+                    if isinstance(v, ForceArgs) and not v.forced_kwargs:
+                        v = v.__wrapped__
+                        setattr(module, k, v)
```
david-berthelot, line 99:
```suggestion
                        setattr(module, k, v.__wrapped__)
```

david-berthelot (changes requested):
Looks very good, only suggested some minor typing hints and code simplification.

objax/module.py
```diff
@@ -53,6 +54,66 @@ def __call__(self, *args, **kwargs):
...
 
 
+class ForceArgs(Module):
+    """Forces override of arguments of given module."""
+
+    ANY = namedtuple('Any', ())
+    """Token used in `ForceArgs.undo` to indicate undo of all values of specific argument."""
+
+    @staticmethod
+    def _remove_args(force_args: 'ForceArgs', args_to_remove: Dict[str, Any]):
```
david-berthelot, line 63: I just realized this should probably not be a static method.
```suggestion
    def remove_args(self, args_to_remove: Dict[str, Any]):
```

objax/module.py
```diff
@@ -53,6 +54,66 @@ def __call__(self, *args, **kwargs):
...
+        """Removes given arguments from override list of `ForceArgs` instance."""
+        if not args_to_remove:
+            force_args.forced_kwargs = {}
+        else:
+            force_args.forced_kwargs = {k: v for k, v in force_args.forced_kwargs.items()
+                                        if (k not in args_to_remove) or (args_to_remove[k] not in (v, ForceArgs.ANY))}
```
david-berthelot, line 69:
```suggestion
            self.forced_kwargs = {}
        else:
            self.forced_kwargs = {k: v for k, v in self.forced_kwargs.items()
                                        if (k not in args_to_remove) or (args_to_remove[k] not in (v, ForceArgs.ANY))}
```

objax/module.py
```diff
@@ -53,6 +54,66 @@ def __call__(self, *args, **kwargs):
...
+
+        Args:
+            module: module for which to undo ForceArgs.
+            **kwargs: dictionary of argument overrides to undo.
+                `name=val` remove override for value `val` of argument `name`.
+                `name=ForceArgs.ANY` remove all overrides of argument `name`.
+                If `**kwargs` is empty then all overrides will be undone.
+        """
+        if isinstance(module, ForceArgs):
+            ForceArgs._remove_args(module, kwargs)
```
david-berthelot, line 83:
```suggestion
            module.remove_args(kwargs)
```

david-berthelot (changes requested):
More suggestions

david-berthelot (approved):

PR #74, opened 2020-09-25
Add signature support to wrappers, add Function module.

### LINKED ISSUE #57: Is it desirable to have an objax.Function module to wrap functions and their vars?
When passing a function to Jit or Grad, etc... we need to also pass the variables the function uses. While we can maintain this design pattern, we could also propose an alternative: a decorator that turns a function into a module.

Example:
```python
# Current design
def some_function(x):
    return objax.functional.softmax(model(x, training=False))

jit_func = objax.Jit(some_function, model.vars())
```

Alternate design proposal (not a replacement)
```python
@objax.Function(model.vars())
def some_function(x):
    return objax.functional.softmax(model(x, training=False))

jit_func = objax.Jit(some_function)
```

I like a certain elegance about this proposal but I would like more opinions on whether it is a desirable addition.

### LINKED ISSUE #58: Make Jit, Vectorize, GradValues compatible with functools.partial
It will probably require using python builtin names like `__wrapped__` and `__signature__`.

david-berthelot:
- Signature support allows good integration with Python `inspect` module (as well as partial). Resolves #58 
- `Function` module is a convenience module to simplify the code for wrappers and can also be used for aesthetic reasons in general purpose user code.  Resolves #57

### DIFF
--- a/objax/module.py
+++ b/objax/module.py
@@ -90,7 +90,10 @@ class Function(Module):
             f: the function or the module to represent.
             vc: the VarCollection of variables used by the function.
         """
-        self.vc = VarCollection((f'{{{f.__name__}}}.{k}', v) for k, v in vc.items())
+        if hasattr(f, '__name__'):
+            self.vc = VarCollection((f'{{{f.__name__}}}.{k}', v) for k, v in vc.items())
+        else:
+            self.vc = VarCollection(vc)
         self.__wrapped__ = f
 
     def __call__(self, *args, **kwargs):
--- a/objax/module.py
+++ b/objax/module.py
@@ -12,9 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__all__ = ['Jit', 'Module', 'ModuleList', 'ModuleWrapper', 'Parallel', 'Vectorize']
+__all__ = ['Function', 'Jit', 'Module', 'ModuleList', 'Parallel', 'Vectorize']
 
-from types import MethodType
 from typing import Optional, List, Union, Callable, Tuple
 
 import jax
@@ -81,26 +80,40 @@ class ModuleList(Module, list):
         return value
 
 
-class ModuleWrapper(Module):
-    """Module whose sole purpose is to store a collectable VarCollection. This class is exclusively
-    used internally by Objax, for example in Jit, Vectorize and Parallel."""
+class Function(Module):
+    """Turn a function into a Module by keeping the vars it uses."""
 
-    def __init__(self, vc: VarCollection):
-        super().__init__()
-        self.vc = VarCollection((f'({self.__class__.__name__}){k}', v) for k, v in vc.items())
-
-    def vars(self, scope: str = '') -> VarCollection:
-        """Collect all the variables (and their names) contained in the VarCollection.
+    def __init__(self, f: Callable, vc: VarCollection):
+        """Function constructor.
 
         Args:
-            scope: string to prefix to the variable names.
-        Returns:
-            A VarCollection of all the variables.
+            f: the function or the module to represent.
+            vc: the VarCollection of variables used by the function.
         """
-        return VarCollection((scope + k, v) for k, v in self.vc.items())
+        self.vc = VarCollection((f'{{{f.__name__}}}.{k}', v) for k, v in vc.items())
+        self.__wrapped__ = f
+
+    def __call__(self, *args, **kwargs):
+        """Call the the function."""
+        return self.__wrapped__(*args, **kwargs)
+
+    def vars(self, scope: str = '') -> VarCollection:
+        """Return the VarCollection of the variables used by the function."""
+        if scope:
+            return VarCollection((scope + k, v) for k, v in self.vc.items())
+        return VarCollection(self.vc)
+
+    @staticmethod
+    def with_vars(vc: VarCollection):
+        """Method to use as decorator in function definitions."""
+
+        def from_function(f: Callable):
+            return Function(f, vc)
+
+        return from_function
 
 
-class Jit(ModuleWrapper):
+class Jit(Module):
     """JIT (Just-In-Time) module takes a function or a module and compiles it for faster execution."""
 
     def __init__(self,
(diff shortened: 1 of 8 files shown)


objax/module.py
```diff
@@ -81,26 +80,40 @@ def __getitem__(self, key: Union[int, slice]):
...
 
         Args:
-            scope: string to prefix to the variable names.
-        Returns:
-            A VarCollection of all the variables.
+            f: the function or the module to represent.
+            vc: the VarCollection of variables used by the function.
         """
-        return VarCollection((scope + k, v) for k, v in self.vc.items())
+        self.vc = VarCollection((f'{{{f.__name__}}}.{k}', v) for k, v in vc.items())
```
AlexeyKurakin, line 93: FIY if `f` is declared as lambda function, then `f.__name__` will be just `<lambda>` which might be not very informative/useful to insert.
So if you have several functions declared as follows:
```
f1 = Function(lambda x, y: ((model(x) - y)**2).sum(), model_vars)
f2 = Function(lambda x, y: ..., model_vars)
```
then naming of the variables in var collection will be the same for both functions.

Another corner case when `f` is declared using partial: `f = functools.partial( .... )`, in such case `f.__name__` is undefined despite f being callable:
```
Python 3.8.5 (default, Jul 20 2020, 18:32:44) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import functools
>>> def f(a, b):
...   return a + b
... 
>>> f.__name__
'f'
>>> f2 = functools.partial(f, b=2)
>>> f2
functools.partial(<function f at 0x7f27596b9430>, b=2)
>>> f2.__name__
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'functools.partial' object has no attribute '__name__'
>>> callable(f2)
True

```

tests/parallel.py
```diff
@@ -15,17 +15,28 @@
...
-
+from objax.typing import JaxArray
 
 # Split CPU cores into 8 devices for tests of objax.Parallel
+
 os.environ['XLA_FLAGS'] = ' '.join(os.environ.get('XLA_FLAGS', '').split(' ')
                                    + ['--xla_force_host_platform_device_count=8'])
 
 
+class LinearArgs(objax.nn.Linear):
```
AlexeyKurakin, line 31: Seems like `LinearArgs` is not used in `parallel.py`

tests/parallel.py
```diff
@@ -15,17 +15,28 @@
...
+
 import jax.numpy as jn
+import numpy as np
 
 import objax
-
+from objax.typing import JaxArray
 
 # Split CPU cores into 8 devices for tests of objax.Parallel
+
```
AlexeyKurakin, line 26: why adding extra blank line?
the comment above directly related to the line `os.environ ....`, in such case usually there is not blank line between comment and the code

tests/jit.py
```diff
@@ -44,6 +68,18 @@ def test_double_jit(self):
...
+    def test_jit_kwargs(self):
+        x = objax.random.normal((64, 3))
+        kj = objax.Jit(LinearArgs(3, 3))
+        y1 = kj(x, 1)
+        y2 = kj(x, some_args=1)
+        y3 = kj(x, some_args=2)
+        self.assertEqual(y1.tolist(), y2.tolist())
+        self.assertNotEqual(y1.tolist(), y3.tolist())
+        kj = objax.Jit(LinearTrain(3, 3))
+        with self.assertRaises(ConcretizationTypeError):
```
AlexeyKurakin, line 80: This error comes from JAX, but maybe it worth some explanation in Objax docs (for example in in-depth guide about Jit). From user point of view, it might be not intuitive why bool argument does not work, while float works.

objax/module.py
david-berthelot: I understand the points you make but not their relevance. What do you suggest?

objax/module.py
```diff
@@ -115,47 +128,31 @@ def __init__(self,
             static_argnums: tuple of indexes of f's input arguments to treat as static (constants)).
                 A new graph is compiled for each different combination of values for such inputs.
         """
-        if vc is None:
-            if not isinstance(f, Module):
+        if not isinstance(f, Module):
```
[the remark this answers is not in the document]
david-berthelot: It's the same behavior as before. I didn't change it. It's already documented in "Understanding gradients...".

tests/jit.py
david-berthelot: I think it's already explained in the docs, in the in-depth guide.

tests/parallel.py
david-berthelot: I changed this file accidentally. Removing all changes.

tests/parallel.py
david-berthelot: I changed this file accidentally. Removing all changes.

objax/module.py
AlexeyKurakin: either don't prepend anything to variable names (because it may not matter that much), or prepend something so any callable will work and ideally different callables could be better differentiated (not sure if it's possible though)

objax/module.py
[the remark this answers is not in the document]
AlexeyKurakin: Maybe add mentioning of this into docstring (or to RST files if docstring will be too long)?
I would expect that API docs is the primary read for users to understand what different arguments do, rather than in-depth tutorials.

objax/module.py
david-berthelot: 1. Not prepending anything: it's less debuggable so there's no upside to it.
2. Prepend so that any callable would work: any callable does work regardless of prepending. It's unrelated.

Basically I prefer to stick to whatever Python does and that's why I designed it this way.

objax/module.py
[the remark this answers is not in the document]
david-berthelot: That would be better suited as a separate PR because this PR is not about changing the behavior of the API.

objax/module.py
AlexeyKurakin: As I mentioned above, `f2 = functools.partial(f1, arg1=val1, ...)` is a callable, but it will cause runtime error because there is not `f2.__name__`.

Basically following code will cause runtime error:

```python
def f1(arg1, arg2):
  # ...

f2 = functools.partial(f1, arg1=val1, ...)
f_module = Function(f2, vars)
```

objax/module.py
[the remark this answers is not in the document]
AlexeyKurakin: It's fine to add in separate PR, I was thinking that it's small enough change of documentation which could be added right here.

objax/module.py
david-berthelot: Thanks, I added code to handle this + test case.

AlexeyKurakin (approved):

PR #75, opened 2020-09-26
Remove type hints.

aterzis-google:
Attempt to remove type hints from readthedocs documentation


david-berthelot (changes requested):
Why are we removing type hints? We want them in the description, don't we?

aterzis-google:
As David mentioned, this is not the way to fix the underlying problem. Closing this request

PR #80, opened 2020-09-30
Allow strings and numerical values for Conv2D/ConvTranspose2D padding.

### LINKED ISSUE #30: Explicit padding mode
It looks like objax currently limits padding to one of VALID or SAME. This prevents the ability to use explicit padding and would prevent compatibility with models from PyTorch, Gluon that only support explicit (symmetric) padding without adding extra Pad layers to the model.

It'd be nice to at minimum add the ability to support TF style explicit padding (specify both sides of every dim), the underlying jax conv impl is able to receive a `[[0, 0], [pad_beg, pad_end],[pad_beg, pad_end], [0, 0]]` spec like other low level TF conv.

Even nicer would be a simplificed, per-spatial dim symmetric values like PyTorch, Gluon `[pad_h, pad_w]` or just `pad` . My default for most 2D convnets in PyTorch is to use `pad = ((stride - 1) + dilation * (kernel_size - 1)) // 2`, which results in a 'same-ish' padding value. This can always be done on top of the full low/high padding sequence above. 

Some TF models explicitly work around the limitations of SAME padding. By limitations, I mean the fact that you end up with input dependent padding that can be aysmmetric and shift your feature maps relative to each other in a manner that varies as you change your input size.
https://github.com/tensorflow/models/blob/146a37c6663e4a249e02d3dff0087b576e3dc3a1/research/deeplab/core/xception.py#L81-L201

Possible interfaces:

* `padding : Union[ConvPadding, Sequence[Tuple[int, int]]]` (like `conv_general_dilated` but with the enum for valid/same)

* Add more modes the enum and associated values for those that need it via a dataclass

```
class PaddingType(enum.Enum):
  """An Enum holding the possible padding values for convolution modules."""
    SAME = 'SAME'
    VALID = 'VALID'
    RAW = 'RAW'  # specify padding as seq of high/low tuples
    SYM = 'SYM'  # specify symmetric padding for spatial dim as tuple for H, W or single int

@dataclass
class Padding:
    type: PaddingType = PaddingType.SAME
    value: Union[Sequence[Tuple[int, int]], Tuple[int, int], int] = None

    @classmethod
    def same(cls):
        return Padding(PaddingType.SAME)

    @classmethod
    def valid(cls):
        return Padding(PaddingType.VALID)

    @classmethod
    def raw(cls, value: Sequence[Tuple[int, int]]):
        return Padding(PaddingType.RAW, value=value)

    @classmethod
    def sym(cls, value: Union[Tuple[int, int], int]):
        return Padding(PaddingType.SYM, value=value)
```

david-berthelot:
Resolves #30

### DIFF
--- a/objax/nn/layers.py
+++ b/objax/nn/layers.py
@@ -25,7 +25,7 @@ from objax import functional, random, util
 from objax.constants import ConvPadding
 from objax.module import ModuleList, Module
 from objax.nn.init import kaiming_normal, xavier_normal
-from objax.typing import JaxArray
+from objax.typing import JaxArray, ConvPaddingInt
 from objax.variable import TrainVar, StateVar
 
 
@@ -145,7 +145,7 @@ class Conv2D(Module):
                  strides: Union[Tuple[int, int], int] = 1,
                  dilations: Union[Tuple[int, int], int] = 1,
                  groups: int = 1,
-                 padding: ConvPadding = ConvPadding.SAME,
+                 padding: Union[ConvPadding, str, ConvPaddingInt] = ConvPadding.SAME,
                  use_bias: bool = True,
                  w_init: Callable = kaiming_normal):
         """Creates a Conv2D module instance.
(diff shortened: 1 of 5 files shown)


kihyuks (approved):
LGTM

PR #85, opened 2020-10-02
Naming manipulation tools to facilitate importing model weights from …

### LINKED ISSUE #42: More control over var/module namespace.

david-berthelot:
…other frameworks and to remap weights between models.

- VarCollection has a new rename method to rename its variables.
- load_var_collection has new option to rename variables from the data read from the file.
- Renamer allows to take a dictionary of strings {source:target}, or a tuple (regex, replacement) or a function (str -> str).

Resolves #42

### DIFF
--- a/docs/source/objax/objax.rst
+++ b/docs/source/objax/objax.rst
@@ -228,7 +228,6 @@ Variables
    :inherited-members:
 
 .. autoclass:: VarCollection
-   :members:
 
     Usage example::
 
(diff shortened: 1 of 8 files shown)


AlexeyKurakin:
Looks good.
Could you add test for `load_var_collection` with renamer?

AlexeyKurakin (approved):

Issue #41, opened 2020-09-16
AddOns


sathish-a:
Hi team, First, I'm excited about this project. I was recently playing with Objax and have implemented the learning rate finder and cyclic learning rate in my pet project. Just to know if we are looking to add it as a part of the library or just an example would be sufficient? And let me know if I can contribute anything so that I can work on it?

AlexeyKurakin:
Thanks for the interest and willingness to contribute.

This sounds like fall into category of examples. Maybe into `examples/learning_rate` subdirectory.
We haven't thought about specific location for add-ons, but may add it in the future if there is interest.

Also, do you have any feedback regarding the current structure of `examples/`?

sathish-a:
Thanks for the suggestion @AlexeyKurakin. How about under `examples/super_convergence` ? We can put the learning rate finder, cyclic learning rate, cyclic momentum, etc. under the above-mentioned directory. I'm currently involved in modularizing the functions. Any suggestions you would like to give?

AlexeyKurakin:
I think `examples/super_convergence` is fine. We're going to revamp structure of examples directory at some point, but then we will move your example as well.

Please add README file to the code. Other than that I don't have specific suggestions, after you start pull request we can go from there.

AlexeyKurakin:
@sathish-a I assigned this issue to you, since you're working on adding super_convergence example

AlexeyKurakin:
@sathish-a friendly ping, just wanted to double check if you still working on it?

sathish-a:
Hi @AlexeyKurakin yes. It's not working as expected when there is an ema in the training op. So debugging on that line. I'll submit PR shortly.

david-berthelot:
By the way, don't hesitate to submit a draft pull request to get feedback/help.

sathish-a:
I'm almost done with this. I have modified the existing `optimizers.Momentum` module to accept the momentum value during the training in order to support cyclical momentum. Now everything is working as expected. Can I share the colab link here so that you guys can review it?

david-berthelot:
Sure, you can even share a pull request. Note, tomorrow we're planning to release 1.1.0, is there any chance you could have a PR before it? If not, that's not the end the world, it'll be in next release (probably November).

sathish-a:
[...]

PR #93, opened 2020-10-08
Override default beta params in Adam

sathish-a:
@david-berthelot, #41 I've added changes to the Adam to override the default `beta1` and `beta2` parameters.


david-berthelot (approved):
Thanks!

PR #94, opened 2020-10-08
Initialize layer-norm properly.

peterjliu:
g should be initialized with ones and b with zeros. Looks like an accidental swap. Otherwise, this layer starts out outputting 1 instead of x.


david-berthelot (approved):
Thanks!

PR #95, opened 2020-10-11
Add warning message if restoring from empty path

carlini:
If you're not careful, sometimes you pass an incorrect path to the checkpoint directory.

By default, objax will create a new directory and then when you call .restore(), it will silently do nothing.

This PR makes objax warn the user that nothing is going to happen, so they have some hope at debugging what's going on. As with the other print() in the checkpoint class, it only prints if verbose is enabled.

### DIFF
--- a/objax/io/checkpoint.py
+++ b/objax/io/checkpoint.py
@@ -92,6 +92,8 @@ class Checkpoint:
         if idx is None:
             all_ckpts = glob.glob(os.path.join(self.logdir, self.DIR_NAME, self.FILE_MATCH))
             if not all_ckpts:
+                if self.verbose:
+                    print('WARNING: Empty path provided; not restoring from any checkpoint.')
                 return 0, ''
             idx = self.checkpoint_idx(max(all_ckpts))
         ckpt = os.path.join(self.logdir, self.DIR_NAME, self.FILE_FORMAT % idx)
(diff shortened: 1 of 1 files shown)


objax/io/checkpoint.py
```diff
@@ -92,6 +92,8 @@ def restore(self, vc: VarCollection, idx: Optional[int] = None):
         if idx is None:
             all_ckpts = glob.glob(os.path.join(self.logdir, self.DIR_NAME, self.FILE_MATCH))
             if not all_ckpts:
+                if self.verbose:
+                    print('WARNING: Empty path provided; not restoring from any checkpoint.')
```
AlexeyKurakin, line 96: Better error message would be: "No checkpoint found, skip restoring variables."

AlexeyKurakin (commented):
I think it's actually valid use case to expect that checkpoint directory is empty.
If you make your code robust to failures, then training loop has to start from checkpoint.restore and very first time it's called on step 0 there is no checkpoint available. In this case restore should do nothing.

I think it's useful to warn user about lack of checkpoint in the directory, but text of warning should be updated.

objax/io/checkpoint.py
carlini: Changed.

AlexeyKurakin (approved):

PR #97, opened 2020-10-12
Move RNN to layers.py and make it stateless.

aterzis-google:


objax/nn/layers.py
```diff
@@ -327,6 +327,63 @@ def __call__(self, x: JaxArray) -> JaxArray:
         self.avg.value += (self.avg.value - x) * (self.momentum - 1)
         return self.avg.value
 
+class RNN(Module):
+    """ Recurrent Neural Network (RNN) block."""
```
AlexeyKurakin, line 331:
```suggestion
    """Recurrent Neural Network (RNN) block."""
```
Extra whitespace in the beginning

objax/nn/layers.py
```diff
@@ -327,6 +327,63 @@ def __call__(self, x: JaxArray) -> JaxArray:
         self.avg.value += (self.avg.value - x) * (self.momentum - 1)
         return self.avg.value
 
+class RNN(Module):
```
AlexeyKurakin, line 330: I think the name `RNN` is too generic.
Pretty much any type of recurrent block (LSTM, GRU, ....) could be called RNN.
Is there some better way to call it?

david-berthelot: Also RNN refers to the architecture, not to the cell. Here's what TF/Keras does https://www.tensorflow.org/api_docs/python/tf/compat/v1/nn/rnn_cell/RNNCell

ebrevdo: This is a specific RNN architecture that operates across time (so not a cell).  I would call this something like SimpleRNN; and make sure it replicates keras' SimpleRNN functionality with default arguments:

https://www.tensorflow.org/api_docs/python/tf/keras/layers/SimpleRNN

RNN you could reserve as an object that takes an RNNCell and performs a scan across time.

objax/nn/layers.py
```diff
@@ -327,6 +327,63 @@ def __call__(self, x: JaxArray) -> JaxArray:
...
+        self.activation = activation
+
+        # Hidden layer parameters
+        self.w_xh = TrainVar(w_init((self.num_inputs, self.nstate)))
+        self.w_hh = TrainVar(w_init((self.nstate, self.nstate)))
+        self.b_h = TrainVar(jn.zeros(self.nstate))
+
+        self.output_layer = Linear(self.nstate, self.num_outputs)
+
+    def __call__(self, inputs: JaxArray, only_return_final=False) -> JaxArray:
```
ebrevdo, line 360: suggest adding a get_initial_state method and optional initial_state argument here

objax/nn/layers.py
```diff
@@ -327,6 +327,63 @@ def __call__(self, x: JaxArray) -> JaxArray:
...
+        """Forward pass through RNN.
+
+        Args:
+            inputs: ``JaxArray`` with dimensions ``num_steps, batch_size, vocabulary_size``.
+            only_return_final: return only the last output if ``True``, or all output otherwise.`
+
+        Returns:
+            Output tensor with dimensions ``N * batch_size, vocabulary_size``.
```
ebrevdo, line 368: This should return a tuple containing `(output, final_state)`.

ebrevdo, line 368: is vocabulary_size the right terminology for RNNs?  perhaps you mean `nout` here?

Also why is batch_size included here?  I thought you don't consider batch_size in these layers?

ebrevdo (changes requested):
Needs a unit test.

jli05 (commented):
What the `RNN` stands out by in this lib for me is the code readability and simplicity. Any person can easily extend it.

objax/nn/layers.py
aterzis-google: Done

objax/nn/layers.py
aterzis-google: Changed the name to SimpleRNN

objax/nn/layers.py
aterzis-google: I added an optional initial_state argument to the call() method. 

Can you clarify what the get_initial_state() method would do, considering that the state is initialized during every call() (unless explicitly passed in through the optional argument)?

objax/nn/layers.py
aterzis-google: Done

aterzis-google: Changed vocabulary_size -> nout 

I include batch_size because we can process a batch of input data.

objax/nn/layers.py
```diff
@@ -327,6 +327,63 @@ def __call__(self, x: JaxArray) -> JaxArray:
...
+            w_init: weight initializer for RNN model weights.
+        """
+        self.num_inputs = nin
+        self.num_outputs = nout
+        self.nstate = nstate
+        self.activation = activation
+
+        # Hidden layer parameters
+        self.w_xh = TrainVar(w_init((self.num_inputs, self.nstate)))
+        self.w_hh = TrainVar(w_init((self.nstate, self.nstate)))
```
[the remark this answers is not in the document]
aterzis-google: Added an assertion that num_inputs > 0

objax/nn/layers.py
```diff
@@ -327,6 +327,63 @@ def __call__(self, x: JaxArray) -> JaxArray:
...
+        """
+        outputs = []
+        state = jn.zeros((inputs.shape[1], self.nstate))
+        for x in inputs:
+            state = self.activation(
+                jn.dot(x, self.w_xh.value)
+                + jn.dot(state, self.w_hh.value)
+                + self.b_h.value
+            )
+            y = self.output_layer(state)
```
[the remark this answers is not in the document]
aterzis-google: I opted for having an ``output_layer``

ebrevdo:
I am ooo and will return next week.

objax/nn/layers.py
ebrevdo: there are two reasons to have a `get_initial_state`: One, the caller wants to know if this layer is recurrent, without checking for some general instance type.  Two, the caller wants to know the shapes etc of the state, without running `__call__`.  This is useful for many reasons, like creating buffers for storing state.

objax/nn/layers.py
ebrevdo: @david-berthelot do other layers "know" about batch dimensions?  does this one need to?

objax/nn/layers.py
```diff
@@ -327,6 +327,63 @@ def __call__(self, x: JaxArray) -> JaxArray:
...
+        Returns:
+            Output tensor with dimensions ``N * batch_size, vocabulary_size``.
+            N = 1 if ``only_return_final`` is ``True`` and ``num_steps`` otherwise.
+        """
+        outputs = []
+        state = jn.zeros((inputs.shape[1], self.nstate))
+        for x in inputs:
+            state = self.activation(
+                jn.dot(x, self.w_xh.value)
+                + jn.dot(state, self.w_hh.value)
```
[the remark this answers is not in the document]
ebrevdo: Typically it's more efficient to act on one concatenated `[h, x]`, but depends on the system and sizes.  At some point you can make this an `__init__` mode parameter like Keras does.  For now I'd suggest using the concatenated format.

objax/nn/layers.py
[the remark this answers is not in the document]
ebrevdo: Question why: this is something the user can do themselves after, right?  So is there any purpose to add an output_layer?

objax/nn/layers.py
david-berthelot: Just to clarify, does `get_init_state` really act like a `create_init_state`? Or is there an `init_state` stored inside the instance?

ebrevdo: no; it's a purely functional thing that returns some arrays.

objax/nn/layers.py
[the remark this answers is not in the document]
david-berthelot: I would drop the output layer, that's forcing a decision on the user about what type of output they'd want.

objax/nn/layers.py
```diff
@@ -332,6 +332,76 @@ def __call__(self, x: JaxArray) -> JaxArray:
...
+                + self.b_h.value
+            )
+            y = self.output_layer(state)
+            if not only_return_final:
+                outputs.append(y)
+
+        if only_return_final:
+            return y, state
+        else:
```
david-berthelot, line 401: No need for `else`.

objax/nn/layers.py
```diff
@@ -332,6 +332,76 @@ def __call__(self, x: JaxArray) -> JaxArray:
...
+            return jn.concatenate(outputs, axis=0), state
```
david-berthelot, line 402: Should it be `jn.stack`?

objax/nn/layers.py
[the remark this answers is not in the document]
david-berthelot: Another nit, use `x.dot(y)` rather than `jn.dot(x, y)` since we might as well take advantage of object oriented APIs.

objax/nn/layers.py
```diff
@@ -332,6 +332,76 @@ def __call__(self, x: JaxArray) -> JaxArray:
...
+
+        # Hidden layer parameters
+        self.w_xh = TrainVar(w_init((self.num_inputs, self.nstate)))
+        self.w_hh = TrainVar(w_init((self.nstate, self.nstate)))
+        self.b_h = TrainVar(jn.zeros(self.nstate))
+
+        self.output_layer = Linear(self.nstate, self.num_outputs, w_init=w_init)
+
+    def __call__(self, inputs: JaxArray, initial_state: JaxArray = None,
+                 only_return_final: bool = False) -> Tuple[JaxArray, JaxArray]:
```
david-berthelot, line 369: One argument per line if they don't all fit on one line.

examples/text_generation/shakespeare_rnn.py
```diff
@@ -172,12 +171,11 @@ def predict_char(prefix, num_predicts, model, vocab):
 
 opt = objax.optimizer.Adam(model_vars)
 ema = objax.optimizer.ExponentialMovingAverage(model_vars, momentum=0.999)
-predict = ema.replace_vars(objax.Jit(lambda x: objax.functional.softmax(model(x)), model_vars))
 
 
 def loss(x, label):  # sum(label * log(softmax(logit)))
-    logit = model(x)
-    return objax.functional.loss.cross_entropy_logits(logit, label).mean()
+    logits, _ = model(x)
```
david-berthelot, line 177: `logits = model(x)[0]`

examples/text_generation/shakespeare_rnn.py
```diff
@@ -137,32 +141,27 @@ def load_shakespeare(batch_size, num_steps, token_type):
...
 def predict_char(prefix, num_predicts, model, vocab):
-    model.init_state(batch_size=1)
     outputs = [vocab[prefix[0]]]
     get_input = lambda: one_hot(jn.array([outputs[-1]]).reshape(1, 1), len(vocab))
     for y in prefix[1:]:  # Warmup state with prefix
         model(get_input())
         outputs.append(vocab[y])
     for _ in range(num_predicts):  # Predict num_predicts steps
-        Y = model(get_input())
+        Y, _ = model(get_input())
```
david-berthelot, line 164: 1. Uppercase are for global constants, use lower case identifiers for variables please.
2. Also rather than doing two assigns, the better way is to just assign what you use.
```Y = model(get_input())[0]```

File: examples/text_generation/shakespeare_rnn.py [hunk 13]
```diff
@@ -22,7 +22,11 @@
 
 import objax
 from objax.functional import one_hot
-from objax.zoo.rnn import RNN
+<<<<<<< HEAD:examples/text_generation/shakespeare_rnn.py
+from objax.nn import SimpleRNN
+=======
+from objax.nn import RNN
+>>>>>>> 2c04d4e (Move RNN to layers.py and make it stateless.):examples/rnn/shakespeare.py
```
david-berthelot, line 29: Your commit contains an unresolved merge.

david-berthelot (changes requested):

objax/nn/layers.py
ebrevdo: (from david on another PR: no, layers don't know about ddimensions, so this one shouldn't either.  instead, add a unit test with this object and `Vectorized`)

objax/nn/layers.py
AlexeyKurakin: As far as I understood from some of the Keras code, `get_initial_state` simply returns zero array of appropriate shape (ex: https://github.com/tensorflow/tensorflow/blob/fcc4b966f1265f466e82617020af93670141b009/tensorflow/python/keras/layers/recurrent.py#L1948 )

It's still not very clear how useful it is.
Could you point us to some example of how it's actually used (either in Tensorflow or any other framework)?

To know shape of the state it would be better to just call `rnn_cell_layer.nstate` or maybe have helper method `get_state_shape`.
Using `get_initial_state` as a way to determine whether layer is RNN seems like a little weird. I don't see how `getattr(layer, 'get_initial_state')` is better than `isinstance(layer, RNNCell)`. If there is a need to determine whether layer is RNN cell, I think it's better just to make all RNN cells to inherit from some base class and do `isinstance` check.

PR #98, opened 2020-10-13
Pooling padding.Resolves #91

### LINKED ISSUE #91: Padding for pooling layers
Now Conv2D allows explicit padding, while pooling layers are not. We need consistent padding types for both convolution and pooling layers.

david-berthelot:
Resolves #91

### DIFF
--- a/tests/functional_pooling.py
+++ b/tests/functional_pooling.py
@@ -16,14 +16,14 @@
 
 import unittest
 
-import jax.numpy as jn
+import jax.numpy as  jn
 import numpy as np
 
 import objax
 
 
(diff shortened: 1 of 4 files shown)


tests/functional_pooling.py
```diff
@@ -16,14 +16,14 @@
 
 import unittest
 
-import jax.numpy as jn
+import jax.numpy as  jn
```
david-berthelot, line 19:
```suggestion
import jax.numpy as jn
```

kihyuks (approved):
LGTM

PR #99, opened 2020-10-13
Give better error when passing wrong input shape to convolution

carlini:
If you pass a tensor with the wrong shape to conv, it gives this completely incomprehensible error message:

```
  File "/home/ncarlini/.local/lib/python3.6/site-packages/jax/lax/lax.py", line 542, in conv_general_dilated
    precision=_canonicalize_precision(precision))
  File "/home/ncarlini/.local/lib/python3.6/site-packages/jax/core.py", line 202, in bind
    out_tracer = top_trace.process_primitive(self, tracers, kwargs)
  File "/home/ncarlini/.local/lib/python3.6/site-packages/jax/interpreters/partial_eval.py", line 133, in process_primitive
    return self.default_process_primitive(primitive, tracers, params)
  File "/home/ncarlini/.local/lib/python3.6/site-packages/jax/interpreters/partial_eval.py", line 141, in default_process_primitive
    out_aval = primitive.abstract_eval(*avals, **params)
  File "/home/ncarlini/.local/lib/python3.6/site-packages/jax/lax/lax.py", line 1672, in standard_abstract_eval
    return ShapedArray(shape_rule(*args, **kwargs), dtype_rule(*args, **kwargs))
  File "/home/ncarlini/.local/lib/python3.6/site-packages/jax/lax/lax.py", line 2167, in _conv_general_dilated_shape_rule
    rhs.shape[dimension_numbers.rhs_spec[1]]))
ValueError: conv_general_dilated lhs feature dimension size divided by feature_group_count must equal the rhs input feature dimension size, but 64 // 1 != 6.
```

This changes it to say something sane
```
  File "/home/ncarlini/objax/objax/nn/layers.py", line 345, in __call__                                                                                                                                                               
    args = f(*args, **util.local_kwargs(kwargs, f))                                                                                                                                                                                   
  File "/home/ncarlini/objax/objax/nn/layers.py", line 183, in __call__                                                                                                                                                               
    self.w.value.shape, x.shape)                                                                                                                                                                                                      
AssertionError: Attempting to convolve an input with 64 input channels when the convolution expects 6 channels. For reference, self.w.shape=(3, 3, 6, 16) and x.shape=(5000, 64, 32, 3).
```

### DIFF
--- a/objax/nn/layers.py
+++ b/objax/nn/layers.py
@@ -166,6 +166,7 @@ class Conv2D(Module):
         super().__init__()
         assert nin % groups == 0, 'nin should be divisible by groups'
         assert nout % groups == 0, 'nout should be divisible by groups'
+        self.nin = nin
         self.b = TrainVar(jn.zeros((nout, 1, 1))) if use_bias else None
         self.w = TrainVar(w_init((*util.to_tuple(k, 2), nin // groups, nout)))  # HWIO
         self.padding = util.to_padding(padding, 2)
@@ -175,6 +176,11 @@ class Conv2D(Module):
 
     def __call__(self, x: JaxArray) -> JaxArray:
         """Returns the results of applying the convolution to input x."""
+        assert x.shape[-1]%self.nin == 0, \
+            'Attempting to convolve an input with {} input channels ' \
+            'when the convolution expects {} channels.' \
+            'For reference, self.w.shape={} and x.shape={}.'.format(x.shape[1], self.nin, 
+                                                                    self.w.value.shape, x.shape)
         y = lax.conv_general_dilated(x, self.w.value, self.strides, self.padding,
                                      rhs_dilation=self.dilations,
                                      feature_group_count=self.groups,
--- a/objax/nn/layers.py
+++ b/objax/nn/layers.py
@@ -176,11 +176,9 @@ class Conv2D(Module):
 
     def __call__(self, x: JaxArray) -> JaxArray:
         """Returns the results of applying the convolution to input x."""
-        assert x.shape[1]%self.nin == 0, \
-            'Attempting to convolve an input with {} input channels ' \
-            'when the convolution expects {} channels.' \
-            'For reference, self.w.shape={} and x.shape={}.'.format(x.shape[1], self.nin, 
-                                                                    self.w.value.shape, x.shape)
+        assert x.shape[1] == self.nin, f'Attempting to convolve an input with {x.shape[1]} input channels ' \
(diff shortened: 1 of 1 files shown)


objax/nn/layers.py
```diff
@@ -175,6 +176,11 @@ def __init__(self,
 
     def __call__(self, x: JaxArray) -> JaxArray:
         """Returns the results of applying the convolution to input x."""
+        assert x.shape[1]%self.nin == 0, \
+            'Attempting to convolve an input with {} input channels ' \
+            'when the convolution expects {} channels.' \
+            'For reference, self.w.shape={} and x.shape={}.'.format(x.shape[1], self.nin, 
+                                                                    self.w.value.shape, x.shape)
```
david-berthelot, line 183: Please use f-strings and proper spacing and also I believe the check should be `a == b` rather than `a % b == 0`, also you refer to `self.w.shape`, I assume you really meant `self.w.value.shape`:
```suggestion
        assert x.shape[1] == self.nin, f'Attempting to convolve an input with {x.shape[1]} input channels ' \
                                   f'when the convolution expects {self.nin} channels.' \
                                   f'For reference, self.w.value.shape={self.w.value.shape} and x.shape={x.shape}.'
```

objax/nn/layers.py
```diff
@@ -166,6 +166,7 @@ def __init__(self,
         super().__init__()
         assert nin % groups == 0, 'nin should be divisible by groups'
         assert nout % groups == 0, 'nout should be divisible by groups'
+        self.nin = nin
```
david-berthelot, line 169: I am not sure we need to add `self.nin` when it can be inferred from `self.w`.

E.g.
```python
@property
def nin(self):
    return self.w.value.shape[-2]
```

What do you think?

david-berthelot (changes requested):
Some code edits, and also I am thinking on whether or not we should make a check function since exactly the same assert could be made for `ConvTranspose2D`.

objax/nn/layers.py
carlini: Is there a reason this is better? A convolution should never hav w.value change shape so I don't think nin should ever go stale. It's more lines of code and uses property magic that I'd avoid by default unless there's a reason you think nin might sometimes be wrong?

david-berthelot: The only reason is coherency and I am not sure about it. Let's just keep `self.nin = nin` since it's a lot simpler as you pointed out.

david-berthelot (approved):

PR #101, opened 2020-10-14
Give better error message when calling Parallel() without replicate()

carlini:
Currently if you forget to call `replicate()` on a `Parallel` module, it dies somewhere in JaX land in between the 5th and 6th circles of hell. This error makes it possible to understand what's going on and find your way back.

### DIFF
--- a/objax/io/checkpoint.py
+++ b/objax/io/checkpoint.py
@@ -92,6 +92,8 @@ class Checkpoint:
         if idx is None:
             all_ckpts = glob.glob(os.path.join(self.logdir, self.DIR_NAME, self.FILE_MATCH))
             if not all_ckpts:
+                if self.verbose:
+                    print('WARNING: Empty path provided; not restoring from any checkpoint.')
                 return 0, ''
             idx = self.checkpoint_idx(max(all_ckpts))
         ckpt = os.path.join(self.logdir, self.DIR_NAME, self.FILE_FORMAT % idx)
--- a/objax/io/checkpoint.py
+++ b/objax/io/checkpoint.py
@@ -93,7 +93,7 @@ class Checkpoint:
             all_ckpts = glob.glob(os.path.join(self.logdir, self.DIR_NAME, self.FILE_MATCH))
             if not all_ckpts:
                 if self.verbose:
-                    print('WARNING: Empty path provided; not restoring from any checkpoint.')
+                    print('No checkpoints found. Skipping restoring variables.')
                 return 0, ''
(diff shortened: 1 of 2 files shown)


carlini:
I see, I didn't quite understand how JaX tracing works. It looks like when you call `Jit()` on a `Parallel` module, it passes a `Tracer` which obviously isn't a Sharded array. One potential fix is to special-case allow the `Jit` to do this. @david-berthelot, do you have thoughts?

david-berthelot (commented):
> I see, I didn't quite understand how JaX tracing works. It looks like when you call `Jit()` on a `Parallel` module, it passes a `Tracer` which obviously isn't a Sharded array. One potential fix is to special-case allow the `Jit` to do this. @david-berthelot, do you have thoughts?

1. Generally speaking don't put a Parallel call inside Jit. (Parallel is a bit like a Jit already)
2. For your question I would simply do something like `isinstance(v, [ShardedDeviceArray, Tracer...])` (not sure what exactly the `Tracer...` class full name is.

carlini:
The reason I found this bug is one of the tests calls Jit on Parallel. Maybe we should fix that too then.

david-berthelot:
> The reason I found this bug is one of the tests calls Jit on Parallel. Maybe we should fix that too then.

Yes, it's just I don't know how to fix it, apart from checking the module passed to Parallel is `isinstance(module, objax.Jit)` but that doesn't appear robust (since one could pass a function instead). That being said, even if it's not perfect, if it catches 90% of user errors that would be great anyway.

david-berthelot (dismissed):

carlini:
I can't reproduce this test case failing. When I run it on my machine it passes. @david-berthelot could you try running this on your machine and see if it works too?

david-berthelot:
It seems reproducible on my machine
```python
import jax, jaxlib
jax.__version__
# '0.2.1'
jaxlib.__version__
# '0.1.55'
```

That's what I got:
```bash
CUDA_VISIBLE_DEVICES= python3 -m unittest tests/parallel.py

# FAIL: test_jit_parallel_bntrain_concat (tests.parallel.TestParallel)
# FAIL: test_jit_parallel_reducers (tests.parallel.TestParallel)
# Ran 12 tests in 8.739s
# 
# FAILED (failures=2)
```

carlini:
JaX version did it. It works on 0.1.63

carlini:
Jax added a new type of tracer; pushed a simple fix.

david-berthelot (approved):

PR #102, opened 2020-10-14
Modify ResNet_V2 implementation to be compatible with Keras implementation.

kihyuks:
Checked compatibility for ResNet50V2, 101V2, 152V2, where pretrained models are available in Keras.


kihyuks:
Will reopen later with load_pretrained_model_weights function

objax/zoo/resnet_v2.py
```diff
@@ -335,6 +367,91 @@ def __init__(self,
...
+    """
+    model_registry = {'ResNet50': {'num_blocks': [3, 4, 6, 3]},
+                      'ResNet101': {'num_blocks': [3, 4, 23, 3]},
+                      'ResNet152': {'num_blocks': [3, 8, 36, 3]}}
+    assert tf is not None, 'Please install tensorflow dependency to be able to load keras weights.'
+    assert arch in model_registry, f'Model weights does not exist for {arch}.'
+    if num_classes != 1000:
+        include_top = False
+    if include_top:
+        assert num_classes == 1000, 'Number of classes should be 1000 when including top layer.'
```
AlexeyKurakin, line 446: could be rewritten in one line as following:
```python
assert not include_top or num_classes == 1000, 'Number of classes should be 1000 when including top layer.'
```

objax/zoo/resnet_v2.py
```diff
@@ -335,6 +367,91 @@ def __init__(self,
...
+        convert_conv_layer(objax_block.proj_conv, [layer for layer in keras_block if layer.name.endswith('0_conv')][0])
+    convert_conv_layer(objax_block.conv_0, [layer for layer in keras_block if layer.name.endswith('1_conv')][0])
+    convert_conv_layer(objax_block.conv_1, [layer for layer in keras_block if layer.name.endswith('2_conv')][0])
+    convert_conv_layer(objax_block.conv_2, [layer for layer in keras_block if layer.name.endswith('3_conv')][0])
+    convert_bn_layer(objax_block.norm_0, [layer for layer in keras_block if layer.name.endswith('preact_bn')][0])
+    convert_bn_layer(objax_block.norm_1, [layer for layer in keras_block if layer.name.endswith('1_bn')][0])
+    convert_bn_layer(objax_block.norm_2, [layer for layer in keras_block if layer.name.endswith('2_bn')][0])
+
+
+def convert_model(model_keras, model_objax, num_blocks: list, include_top: bool = True):
```
AlexeyKurakin, line 423: maybe rename to `convert_keras_model`

objax/zoo/resnet_v2.py
```diff
@@ -335,6 +367,91 @@ def __init__(self,
...
+    """Convert keras implementation of ResNetV2 into objax."""
```
AlexeyKurakin, line 424:
```suggestion
    """Convert Keras implementation of ResNetV2 into Objax."""
```

objax/zoo/resnet_v2.py
```diff
@@ -335,6 +367,91 @@ def __init__(self,
...
+    convert_conv_layer(model_objax[0], model_keras.get_layer('conv1_conv'))
+    for b, j in enumerate(num_blocks):
+        for i in range(j):
+            convert_resblock(model_objax[b + 3][i], [layer for layer in model_keras.layers if layer.name.startswith(
```
AlexeyKurakin, line 428: Line wrapping looks weird, either extract `[layer for layer in ...]` into separate line (by assigning it to the variable) or change line wrapping, for example:

```python
convert_resblock(model_objax[b + 3][i],
                 [layer for layer in model_keras.layers
                  if layer.name.startswith('conv{}_block{}'.format(b + 2, i + 1))])
```

tests/resnet_v2.py
```diff
@@ -43,5 +47,49 @@ def test_resnet(self, name, resnet_cls):
...
+        sq_diff = (output_objax - output_keras) ** 2
+        self.assertAlmostEqual(sq_diff.mean(), 0, delta=1.001e-07)
+
+    def check_output_shape(self, model, num_classes):
+        data = np.random.uniform(size=(2, 3, 224, 224))
+        output = model(data, training=False)
+        self.assertEqual(output.shape, (2, num_classes))
+
+    def test_resnet_include_top(self):
+        for arch in ['ResNet50', 'ResNet101', 'ResNet152']:
```
AlexeyKurakin, line 75: use `@parameterized`, see example above in this file

AlexeyKurakin (changes requested):

objax/zoo/resnet_v2.py
```diff
@@ -335,6 +367,91 @@ def __init__(self,
...
+
+def load_pretrained_weights_from_keras(arch: str, include_top: bool = True, num_classes: int = 1000):
+    """Function to load weights from keras models.
+    """
+    model_registry = {'ResNet50': {'num_blocks': [3, 4, 6, 3]},
+                      'ResNet101': {'num_blocks': [3, 4, 23, 3]},
+                      'ResNet152': {'num_blocks': [3, 8, 36, 3]}}
+    assert tf is not None, 'Please install tensorflow dependency to be able to load keras weights.'
+    assert arch in model_registry, f'Model weights does not exist for {arch}.'
+    if num_classes != 1000:
```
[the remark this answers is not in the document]
kihyuks: How about we remove this and add the following
```
include_top = include_top and num_classes == 1000
```
When num_classes != 1000, include_top should be set False. I thought this may be better to be auto-corrected since user's intention is clear when using num_classes to be different from 1000. However, it also make sense to throw an error to warn user if configurations are not compatible though. WDYT?

AlexeyKurakin:
@kihyuks is it ready for review?
I saw that some new changes were commited, but seems like most of the comments were not addressed

rwightman:
@kihyuks A question related to your work on this, how close are you validation results on these models after loading keras weights? 

I've been bashing my head trying to figure out if I made a mistake or there's something more fundamental w/ JAX or Objax I'm missing. I have been working with a different model family port, loading Tensorflow or PyTorch weights and find I'm > 1% top-1 validation difference despite accounting for padding, BN eps, and the input pipeline essentially being the same as original. 

That's much more than I've ever seen moving between Tensorflow, PyTorch, MXNet in the past AND changing input pipelines (different impl of bilinear/bicubic interoplation, etc).

kihyuks:
@AlexeyKurakin sorry, I just pushed changes. Please take another look.

objax/zoo/resnet_v2.py
```diff
@@ -335,6 +356,98 @@ def __init__(self,
...
+    """Converts variables of convolution layer from Keras to Objax."""
+    objax_layer.w = objax.TrainVar(jn.array(keras_layer.__dict__['kernel'].numpy()))
+    if keras_layer.__dict__.get('bias') is not None:
+        bias_shape = (objax_layer.w.value.shape[-1], 1, 1)
+        objax_layer.b = objax.TrainVar(jn.array(keras_layer.__dict__['bias'].numpy()).reshape(bias_shape))
+    else:
+        objax_layer.b = None
+
+
+def convert_linear_layer(keras_layer, objax_layer):
```
AlexeyKurakin, line 384: `convert_linear_layer` and `convert_conv_layer` are almost identical, except for the line which computes `bias_shape`.
However for `bias_shape` I think you can just use code `bias_shape = objax_layer.b.value.shape`.

So I think it's better to have one function, instead of two identical functions. You can call it `convert_conv_or_linear_layer`

AlexeyKurakin (commented):
Just few minor comments.

Also please add `load_pretrained_weights_from_keras` to https://github.com/google/objax/blob/master/docs/source/objax/zoo.rst
so documentation would be generated for this function.

objax/zoo/resnet_v2.py
kihyuks: Done. However, `bias_shape = objax_layer.b.value.shape` didn't work because currently objax conv layers are all defined with `use_bias=False`, while some keras layers use biases.

kihyuks:
@AlexeyKurakin PTAL.

AlexeyKurakin (approved):

PR #108, opened 2020-10-15
Add CIFAR10 Colab tutorial

AlexeyKurakin:


aterzis-google (commented):
My main comment is that this implementation of CIFAR10 classification is very much like https://github.com/google/objax/blob/master/examples/classify/img/cifar10_simple.py but not exactly (e.g., the chosen optimizers are different). 

Should we try to make the two as similar as possible to avoid questions about why they are different?

AlexeyKurakin:
Initially I indeed based this colab on cifar10_simple, however I wanted to make it more in-line with WideResnet paper and increase accuracy compared to cifar10_simple.

I added explanation of differences between cifar10_simple and this colab.

I had impression that cifar10_simple was intentionally kept over-simplified. This colab adds few features which let it achieve higher accuracy (>=93%).

AlexeyKurakin:
If you think it's better to keep these two examples as similar as possible then I think I would probably prefer to update cifar10_simple.py example to match this colab

aterzis-google (approved):
I think the explanation is adequate

Issue #106, opened 2020-10-15
Add type and shape check to var assign


rwightman:
I made a few mistakes moving weights from PyTorch with vc.assign and ended up clobbering over all of the models weights with the wrong shape. No errors until you try to use the model.

I noticed the assign fn just uses `var = tensor` ... no copy option since the jax array's are immutable, but wouldn't an `assert isinstance(tensor, JaxArray) and self.var.shape == tensor.shape` be appropriate? Or possibly an attempt to convert the type to JaxArray and broadcast the shape... 

Am I missing use cases where you'd want to change type away from JaxArray or use a different shape than the original on assign()?

david-berthelot:
We started an effort on creating better error messages (since the JAX ones can lead to confusion) which include adding asserts in various places. Your proposal fits well into that project: https://github.com/google/objax/projects/1

1. Assert `isinstance` seems like a good idea.
2. Assert `shape` also seems like a good idea since I don't see under what scenario modifying a variable value to one with a different shape would be useful. In that case, I would expect someone to replace a variable entirely rather than replacing its value. That being said, I'd love some feedback in case anyone can think of a scenario where that would make sense.

carlini:
This bug just bit me when accidentally restoring a checkpoint with different sized models. For example, if you have WRN-18,2 and restore a WRN-18,1 it'll clobber over the prior sizes with the new sizes and then you get very weird behavior.

I'm writing a PR to address both of these bugs.

carlini:
Fixed in #123.

PR #115, opened 2020-10-18
Give error if assigning variable wrong shape

carlini:
This PR asserts that the shape of a variable isn't changed when assigning to it. Most of it is fairly uncontroversial (#106).

However I do one further thing. In order for the error message when restoring a variable show the name of the variable that caused the error, I add some additional state to BaseVar to support this. As a result, if restore fails, it won't just say that *some* variable had the wrong shape, but now it will error with exactly *which* variable did.

The other way to do this would be to wrap the restore assign() in a try/catch and then repeat the error, but with the name attached. The benefit of this approach is the error isn't duplicated, but it's a bit more code.

(I also imagine that having the variable name could help for some other errors in the future.)

### DIFF
--- a/objax/io/ops.py
+++ b/objax/io/ops.py
@@ -58,9 +58,11 @@ def load_var_collection(file: Union[str, IO[BinaryIO]],
         for name in names:
             index = name_index.get(name)
             if index is not None:
-                v._name = name # make the debugging easier if the assign fails
-                v.assign(jn.array(data[index]))
-                v._name = ""
+                try:
+                    v.assign(jn.array(data[index]))
+                except e:
+                    e.message = f"Error when restoring variable {name}: " + message
+                    raise
                 break
         else:
             misses += names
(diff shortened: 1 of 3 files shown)


carlini:
Ah well. This is interesting, and is going to have to be a bigger PR than the trivial change I'd thought:

1. Replicate() assigns a variable of a different shape, but this needs to be allowed.

2. Some of the other tests abuse this feature and need to be fixed.

carlini:
It looks like the test has stalled (or is taking >12 hours)

david-berthelot (changes requested):
Seems some changes came from another PR (https://github.com/google/objax/pull/101) for `module.py`, resulting in conflicting files since they were updated in the other PR but not here. So the conflicts need to be resolved.

Generally it's better to keep changes local to each PR, was it an accident?

carlini:
Huh. I'm not sure how I got in that state, but it wasn't intended.

carlini:
Ah well. I've just completely messed up my local state. I'm going to close this PR and re-open one with a clean history.

PR #120, opened 2020-10-20
Examples cleanup

schien1729:
**Current Issues:**

- Very high (number of directories / number of code files) ratio
- Organization is inconsistent: top-level directories are variously named for type of task (classify), type of architecture (rnn), or model itself (gpt-2).
- Within classify directory, inconsistency in directory structure (img/ has an imagenet subdirectory, but everything else (mnist, cifar) is at top level, including another imagenet model.
- A number of utilities (I think) are not documented with their purpose.
- classify also has a semi_supervised subdirectory. Unless all semi-supervised examples are going to be classification examples, it should probably not be here.
- There are a lot of directories with only one subdirectory.

**New Guidelines**

- Fewer dirs with only one subdir.
- Separate folders for common tutorial tasks (images, text)
- Separate folders for specific research work (one folder per paper); open to future consolidation if there are several similar papers


examples/README.md
```diff
@@ -2,9 +2,16 @@
...
-See:
-* [Image Classification](classify/img/README.md)
+Examples from classic machine learning:
+* [Image Classification](image_classification/README.md)
+* [Text Generation](text_generation/README.md)
+
+Examples from recent research:
 * [Model-Agnostic Meta-Learning](optimization/README.md)
-* [Recurrent Neural Networks (RNN)](rnn/README.md)
+* [FixMatch](fixmatch/README.md)
```
AlexeyKurakin, line 13: ** Comments about fixmatch move:

1) it seems like it moved to `examples/fixmatch/img/`. However right now we only have Fixmatch for images and it's unlikely that we will have fixmatch for other domains in any near future. This this Fixmatch should be moved to `examples/fixmatch`

2) You need to update all imports in fixmatch code.
It seems like right now fixmatch files are moved to new location, however they still use old imports like `from examples.classify.semi_supervised.img... import ...`

3) Also fixmatch path need to be updated in the `example/fixmatch/README.md`

examples/README.md
```diff
@@ -2,9 +2,16 @@
...
+
+Other examples:
+* [Tutorials](tutorials/README.md)
```
AlexeyKurakin, line 16: There is one more file `cifar10.ipynb` recently added in `tutorials`, which is not moved in this pull request

examples/README.md
```diff
@@ -2,9 +2,16 @@
...
+This directory contains fully functional code examples that you can use to learn more about Objax.
 
-See:
-* [Image Classification](classify/img/README.md)
+Examples from classic machine learning:
+* [Image Classification](image_classification/README.md)
+* [Text Generation](text_generation/README.md)
+
+Examples from recent research:
 * [Model-Agnostic Meta-Learning](optimization/README.md)
```
AlexeyKurakin, line 12: I think you renamed `optimization` directory into `maml`.
Please update link here as well

examples/README.md
```diff
@@ -2,9 +2,16 @@
...
-* [Recurrent Neural Networks (RNN)](rnn/README.md)
+* [FixMatch](fixmatch/README.md)
+
```
AlexeyKurakin, line 14: Please add reference to gpt-2 subdirectory

examples/README.md
```diff
@@ -2,9 +2,16 @@
 
 # Examples
 
-This directory contains fully functional code examples that you can use to learn more about Objax
+This directory contains fully functional code examples that you can use to learn more about Objax.
 
-See:
-* [Image Classification](classify/img/README.md)
+Examples from classic machine learning:
+* [Image Classification](image_classification/README.md)
```
AlexeyKurakin, line 8: In image classification, I think `horses_logistic.py` should be renamed.
I think it would be better either to remove name of the class, i.e. use `logistic.py` or `logistic_regression.py`
Or include names of both classes. i.e. `horses_or_humans_logistic.py`

For all other files we use dataset name (`mnist_dnn.py`, `imagenet_...`), which is inconsistent with `horses_logistic.py`

examples/image_classification/README.md
```diff
@@ -27,8 +27,12 @@ This directory contains various classification examples on image datasets:
 
 * [Imagenet](http://www.image-net.org/challenges/LSVRC/2012/)
 
-  * `pretrained_vgg.py` - example which shows how to load pre-trained weights for a VGG model and use it
+  * `imagenet_pretrained_vgg.py` - example which shows how to load pre-trained weights for a VGG model and use it
     to classify input images. For more details see [documentation](pretrained_vgg.md).
 
-  * `imagenet/imagenet_train.py` - example which shows how to train Resnet50 model on Imagenet.
+  * `imagenet_resnet50_train.py` - example which shows how to train Resnet50 model on Imagenet.
     For more details see example [documentation](imagenet/README.md).
```
AlexeyKurakin, line 34: please fix links to documentation for imagenet examples, it should be `imagenet_pretrained_vgg.md` and `imagenet_resnet50.md`

examples/image_classification/cifar10_advanced.py
```diff
@@ -25,7 +25,7 @@
 from tqdm import tqdm, trange
 
 import objax
-from examples.classify.img.tfdata.data import DataSet
+from examples.image_classification.tfdata.data import DataSet
```
AlexeyKurakin, line 28: Ideally I think we should remove `examples.image_classification.tfdata.data` completely and copy content of `DataSet` directly into this file.
The reason is because `examples.image_classification.tfdata.data` is only used in cifar10_advanced example, moreover it contains bunch of unused stuff. Thus presence of examples/image_classification/tfdata/data.py might add confusion for users

But it might be too much change for this pull request, so we can defer it for later.

AlexeyKurakin (changes requested):
It seems like I can not add comments for files which were moved without changing content.
So I wrote all my comments about file renames into examples/README.md

examples/README.md
schien1729: Done.

examples/README.md
schien1729: Done.

examples/README.md
schien1729: Done.

examples/README.md
schien1729: Done.

examples/README.md
schien1729: Done.

examples/image_classification/README.md
schien1729: Done.

examples/image_classification/cifar10_advanced.py
schien1729: Agree in both principle and decision to leave for later.

schien1729:
Hi Alex-- Thanks for the comments. Please take another look whenever is convenient!

AlexeyKurakin (approved):

PR #121, opened 2020-10-21
Remove BaseState limitations in JAX wrapping.

### LINKED ISSUE #114: "objax.variable.VarCollection is not a valid JAX type" when creating a custom optimizer
Hi, I wish to create a custom optimizer to replace the 
`opt(lr=lr, grads=g)
`
line in the example https://github.com/google/objax/blob/master/examples/classify/img/cifar10_simple.py

Instead, I replaced it with 
```
for grad, p in zip(g, model_vars):
      p.value -= lr * grad   
```
and then supplied model.vars() as an argument to train_op. However, I received an error: objax.variable.VarCollection is not a valid JAX type. Can someone help me with this issue? Here is a minimal working example which reproduces the error. 

```
import random
import numpy as np
import tensorflow as tf
from objax.zoo.wide_resnet import WideResNet

# Data
(X_train, Y_train), (X_test, Y_test) = tf.keras.datasets.cifar10.load_data()
X_train = X_train.transpose(0, 3, 1, 2) / 255.0
X_test = X_test.transpose(0, 3, 1, 2) / 255.0

# Model
model = WideResNet(nin=3, nclass=10, depth=28, width=2)
#opt = objax.optimizer.Adam(model.vars())
predict = objax.Jit(lambda x: objax.functional.softmax(model(x, training=False)),
                    model.vars())
# Losses
def loss(x, label):
    logit = model(x, training=True)
    return objax.functional.loss.cross_entropy_logits_sparse(logit, label).mean()

gv = objax.GradValues(loss, model.vars())

def train_op(x, y, model_vars, lr):
    g, v = gv(x, y)
    for grad, p in zip(g, model_vars):
      p.value -= lr * grad   
    return v


# gv.vars() contains the model variables.
train_op = objax.Jit(train_op, gv.vars()) #I deleted opt.vars()

for epoch in range(30):
    # Train
    loss = []
    sel = np.arange(len(X_train))
    np.random.shuffle(sel)
    for it in range(0, X_train.shape[0], 64):
        loss.append(train_op(X_train[sel[it:it + 64]], Y_train[sel[it:it + 64]].flatten(), model.vars(), 4e-3 if epoch < 20 else 4e-4)) #I added model.vars() 
```

david-berthelot:
Resolves #114


tests/parallel.py
```diff
@@ -271,6 +271,18 @@ def test_jit_parallel_reducers(self):
...
+        m = objax.ModuleList([objax.TrainVar(jn.zeros(2))])
+
+        def increase():
+            m[0].assign(m[0].value + 1)
+            return m[0].value
+
+        para_increase = objax.Parallel(increase, m.vars())
+        with m.vars().replicate():
+            para_increase()
+        self.assertEqual(m[0].value.tolist(), [1., 1.])
```
AlexeyKurakin, line 284: Maybe add another test which does increase by different values in different replicas:
```

def increase(x):
  m[0].assign(m[0].value + x)
  return m[0].value

x = np.arange(8)
para_increase = objax.Parallel(increase, m.vars())
with m.vars().replicate():
  para_increase()

...

```

tests/vectorize.py
```diff
@@ -138,6 +139,19 @@ def __call__(self, x):
...
+
+        def increase(x):
+            m[0].assign(m[0].value + 1)
+            return x + 1
+
+        x = np.arange(10)[:, None]
+        vec_increase = objax.Vectorize(increase, m.vars())
+        y = vec_increase(x)
+        self.assertEqual(y.tolist(), np.arange(1, 11)[:, None].tolist())
+        self.assertEqual(m[0].value.tolist(), [1., 1.])
```
AlexeyKurakin, line 153: Similar to parallel, add another test which tests a situation when TrainVar is increased by different values (in other words test that reduce will work correctly in this case):

```
        def increase(x):
            m[0].assign(m[0].value + x)
            return x + 1

   ...
```

AlexeyKurakin (commented):
Few cases which I think worth including in the tests:

1. When variables are updated by optimizer and via `TrainVar.assign` at the same time (i.e. inside jitted function).
I would expect that this should work because TrainRef contains a reference to TrainVar, so ultimately the same TrainVar is getting updated.
However this is still interesting because in this case `changes` will return both TrainVar and TrainRef

2. `Jit` of `Grad`, when variable assignment happens inside `Grad` and then separately inside Jit

3. In both `Vectorize` and `Parallel` test what would happen when different replicas assign different values to the TrainVar (see inline comments in the code)

tests/parallel.py
david-berthelot: Done

tests/vectorize.py
david-berthelot: Done

AlexeyKurakin (approved):

PR #123, opened 2020-10-21
Raise error on changing shape in assign

carlini:
This PR is a cleaned up version of #115.

The checking logic is somewhat complicated in case someone decides to JIT a Parallel() module, it will still do the right thing and won't accidentally cause errors.


objax/io/ops.py
```diff
@@ -60,7 +60,10 @@ def load_var_collection(file: Union[str, IO[BinaryIO]],
             index = name_index.get(name)
             if index is not None:
                 used_vars.add(name)
-                v.assign(jn.array(data[index]))
+                try:
+                    v.assign(jn.array(data[index]))
+                except AssertionError as e:
+                    raise AssertionError(f"Error when restoring variable {name}: " + str(e)) from None
```
david-berthelot, line 66:
```suggestion
                    raise AssertionError(f'Error when restoring variable {name}: {str(e)}') from None
```

objax/variable.py
```diff
@@ -53,15 +53,44 @@ def value(self) -> JaxArray:
...
         self.value = tensor
 
     def reduce(self, tensors: JaxArray):
         """Method called by Parallel and Vectorize to reduce a multiple-device (or batched in case of vectoriaation)
         value to a single device."""
         if self._reduce:
-            self.assign(self._reduce(tensors))
+            self.assign(self._reduce(tensors), check=False)
+
+    def assert_assigned_type_and_shape_match(self, tensor):
```
david-berthelot, line 68: 1. I think we should create a submodule of `objax.util` called `objax.util.check` because this is a lot of checking code.
2. We should name the function `check_assign_type_and_shape(target, source)`

objax/variable.py
```diff
@@ -53,15 +53,44 @@ def value(self) -> JaxArray:
...
+        assert isinstance(tensor, JaxArray.__args__), \
+            f'Assignments to variable must be an instance of JaxArray,  but received f{type(tensor)}.'
+
+        def shape_and_device(array):
```
david-berthelot, line 72: This function should be also defined in `objax.util.check`, defining a function inside a function may incur a runtime penalty.

objax/variable.py
```diff
@@ -53,15 +53,44 @@ def value(self) -> JaxArray:
...
     def reduce(self, tensors: JaxArray):
         """Method called by Parallel and Vectorize to reduce a multiple-device (or batched in case of vectoriaation)
         value to a single device."""
         if self._reduce:
-            self.assign(self._reduce(tensors))
+            self.assign(self._reduce(tensors), check=False)
+
+    def assert_assigned_type_and_shape_match(self, tensor):
+        assert isinstance(tensor, JaxArray.__args__), \
+            f'Assignments to variable must be an instance of JaxArray,  but received f{type(tensor)}.'
```
david-berthelot, line 70: Double space.

objax/variable.py
```diff
@@ -53,15 +53,44 @@ def value(self) -> JaxArray:
...
+            else:
+                return None, array.shape
+        tensor_device, tensor_shape = shape_and_device(tensor)
+        self_device, self_shape = shape_and_device(self.value)
+
+        device_mismatch_error = f"Can not replicate a variable that is currently on " \
+                                f"{self_device} devices to {tensor_device} devices."
+        assert (tensor_device is None) or (self_device is None) or (self_device == tensor_device), device_mismatch_error
+
+        tracer_types = (jax.interpreters.partial_eval.JaxprTracer,
```
david-berthelot, line 84: Define a global constant in `objax.util.check`.

objax/variable.py
```diff
@@ -53,15 +53,44 @@ def value(self) -> JaxArray:
...
+
+        def shape_and_device(array):
+            if isinstance(array, jax.interpreters.pxla.ShardedDeviceArray):
+                return array.shape[0], array.shape[1:]
+            else:
+                return None, array.shape
+        tensor_device, tensor_shape = shape_and_device(tensor)
+        self_device, self_shape = shape_and_device(self.value)
+
+        device_mismatch_error = f"Can not replicate a variable that is currently on " \
```
david-berthelot, line 80: Let's using single quote to be consistent with the rest of the code.

david-berthelot (changes requested):
We also need to add unit tests to make sure the assert catch errors properly and that future commits don't break them.

objax/util/check.py
```diff
@@ -0,0 +1,51 @@
...
+
+    new_tensor_device, new_tensor_shape = split_shape_and_device(new_tensor)
+    self_device, self_shape = split_shape_and_device(existing_tensor)
+
+    device_mismatch_error = f'Can not replicate a variable that is currently on ' \
+                            f'{self_device} devices to {new_tensor_device} devices.'
+    assert (new_tensor_device is None) or (self_device is None) or (self_device == new_tensor_device), \
+        device_mismatch_error
+
+    tracer_types = (jax.interpreters.partial_eval.JaxprTracer,
```
david-berthelot, line 41: Can you make this a global constant, otherwise it recreates a list for each var assignment check.

objax/util/check.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def split_shape_and_device(array):
+    if isinstance(array, jax.interpreters.pxla.ShardedDeviceArray):
+        return array.shape[0], array.shape[1:]
+    else:
+        return None, array.shape
+
+
+def assert_assigned_type_and_shape_match(existing_tensor, new_tensor):
```
david-berthelot, line 29: One concern is this function is super hard to understand. I'm not sure if we that's something we must accept or if it could be made simpler. If we don't think anything simpler at this point we could go with it and revisit in the future if we have new ideasa.

david-berthelot (changes requested):

objax/variable.py
carlini: Done

objax/util/check.py
carlini: I can add some comments to the code to explain why it's doing things. It's really easy to understand except for the jax tracing makes it all very complicated.

I could significantly simplify it by saying that if either type is a jax tracer just return true, but I tried to add some checks to this still.

carlini:
Build fails as a result of #124.

objax/util/check.py
david-berthelot: Well no need for now. Let's just move on with what we have right now.

For the failing test, maybe adjust the tolerance so that it passes.

david-berthelot (approved):

objax/util/check.py
```diff
@@ -0,0 +1,52 @@
...
+# limitations under the License.
+
+__all__ = ['assert_assigned_type_and_shape_match']
+
+import jax
+
+from objax.typing import JaxArray
+
+
+tracer_types = (jax.interpreters.partial_eval.JaxprTracer,
```
david-berthelot, line 22: Global constants should be in upper case.
```suggestion
TRACER_TYPES = (jax.interpreters.partial_eval.JaxprTracer,
```

carlini:
How did this test ever pass. Either that or I'm getting extraordinarily unlucky.

david-berthelot (changes requested):

david-berthelot (approved):

carlini:
@googlebot I consent

carlini:
(sorry for the spam everyone, trying to fix some thing)

PR #136, opened 2020-11-04
LARS optimizer

aterzis-google:


tests/optimizer.py
```diff
@@ -28,11 +28,13 @@ class TestOptimizers(unittest.TestCase):
     def __init__(self, methodname):
         """Initialize the test class."""
         super().__init__(methodname)
-        self.num_steps = 100
+        self.num_steps = 200
```
david-berthelot, line 31: Why make it 200? Is it going to double the runtime of the test?

objax/optimizer/lars.py
```diff
@@ -0,0 +1,67 @@
...
+        clipped_trust_ratio = jn.where(jn.logical_or(grad_norm == 0., train_vars_norm == 0.), 1., trust_ratio)
+        scaled_lr = lr * clipped_trust_ratio
+
+        for g, p, m in zip(grads, self.train_vars, self.m):
+            if self.wd != 0:
+                g += self.wd * p.value
+            scaled_grad = scaled_lr * g
+            new_momentum = self.beta * m.value + scaled_grad
+            p.value -= new_momentum
+            m.value = new_momentum
```
david-berthelot, line 67: Simplify
```suggestion
            m.value = self.beta * m.value + scaled_lr * g
            p.value -= m.value
```

objax/optimizer/lars.py
```diff
@@ -0,0 +1,67 @@
...
+            vc: collection of variables to optimize.
+        """
+        self.beta = beta
+        self.wd = wd
+        self.tc = tc
+        self.eps = eps
+        self.train_vars = ModuleList(TrainRef(x) for x in vc.subset(TrainVar))
+        self.m = ModuleList(StateVar(jn.zeros_like(x.value)) for x in self.train_vars)
+
+    def __call__(self, lr: float, grads: List[JaxArray]):
```
david-berthelot, line 46: Missing type hint for return value.

objax/optimizer/lars.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class LARS(Module):
+    """Layerwise adaptive rate scaling (LARS) optimizer.
+
+    See https://arxiv.org/abs/1708.03888
+    """
+
+    def __init__(self, vc: VarCollection, beta: float = 0.9, wd: float = 0.0001,
+                 tc: float = 0.001, eps: float = 0.00001):
```
david-berthelot, line 33: 1. If declarations don't all fit on one line, use one line per argument.
2. Document arguments.
3. Use 1e-4, 1e-3, 1e-5 otherwise it's hard to known many 0 there are.
4. Documentation in `/docs/` is missing.

objax/optimizer/__init__.py
```diff
@@ -16,3 +16,4 @@
 from .ema import *
 from .momentum import *
 from .sgd import *
+from .lars import *
```
david-berthelot, line 19: Alphabetical order please.

objax/optimizer/lars.py
```diff
@@ -0,0 +1,67 @@
...
+
+from objax.module import Module, ModuleList
+from objax.typing import JaxArray
+from objax.variable import TrainRef, StateVar, TrainVar, VarCollection
+
+
+class LARS(Module):
+    """Layerwise adaptive rate scaling (LARS) optimizer.
+
+    See https://arxiv.org/abs/1708.03888
```
david-berthelot, line 29: After reading the paper, it seems the implementation is not following Algorithm-1. IIC the norm should be layer-wise while in this implementation the norm is network-wise. Or did I miss some subtlety?

david-berthelot (changes requested):

objax/optimizer/lars.py
aterzis-google: There is no return value and we don't use "-> None" in the other optimizers (e.g., objax/optimizer/adam.py)

objax/optimizer/__init__.py
aterzis-google: done

objax/optimizer/lars.py
aterzis-google: Done, thanks!

objax/optimizer/__init__.py
david-berthelot: Hmm, maybe you didn't commit your changes because it doesn't appear to be done.

david-berthelot (changes requested):

tests/optimizer.py
aterzis-google: Switched back to 100 after making the LARS test succeed within 100 steps

objax/optimizer/lars.py
aterzis-google: 1. Done
2. Done
3. Done
4. Done

objax/optimizer/lars.py
aterzis-google: Updated to perform operations on a per-tensor basis

objax/optimizer/__init__.py
aterzis-google: should be in the latest push

objax/optimizer/lars.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+
+    def __init__(self, vc: VarCollection,
+                 momentum: float = 0.9,
+                 beta: float = 1e-4,
```
david-berthelot, line 34: Let's rename this to `weight_decay` since `beta` sounds obscure.

objax/optimizer/lars.py
```diff
@@ -0,0 +1,78 @@
...
+                 max_epochs: int = 100):
+        """Constructor for LARS optimizer.
+
+        Args:
+            vc: collection of variables to optimize.
+            momentum: coefficient used for the moving average of the gradient.
+            beta: weight decay coefficient.
+            tc: trust coefficient eta ( < 1) for trust ratio computation.
+            eps: epsilon used for trust ratio computation.
+            max_epochs: total number of steps that the optimizer will take.
```
david-berthelot, line 46: If I understand well what is called here `max_epoch` is the parameter `T` in the paper.
From the look of it, it's a built-in learning rate decay which should be independent of the method itself for the sake of generality.

What I mean by that is that one could simply invoke `__call__` with `((1-t/T)**2, grads)` and not be tied to a polynomial learning rate decay.

objax/optimizer/lars.py
```diff
@@ -0,0 +1,78 @@
...
+            grads: the gradients to apply.
+        """
+        assert len(grads) == len(self.train_vars), 'Expecting as many gradients as trainable variables'
+        self.epoch += 1
+
+        for g, p, m in zip(grads, self.train_vars, self.m):
+            train_vars_norm = jn.linalg.norm(p.value)
+            grad_norm = jn.linalg.norm(g)
+            trust_ratio = self.tc * train_vars_norm / (grad_norm + self.beta * train_vars_norm + self.eps)
+            clipped_trust_ratio = jn.where(jn.logical_or(grad_norm == 0., train_vars_norm == 0.), 1., trust_ratio)
```
david-berthelot, line 71: I don't understand how `clipped_trust_ratio` relate to the paper. Where does this come from?

objax/optimizer/lars.py
```diff
@@ -0,0 +1,78 @@
...
+            lr *= (1 - self.epoch / self.max_epochs) ** 2
+            scaled_lr = lr * clipped_trust_ratio
+
+            if self.beta != 0:
```
david-berthelot, line 75: Not sure if the `if` statement makes any difference here. A multiply by zero should be typically be optimized away automatically anyway.

david-berthelot (changes requested):

objax/optimizer/lars.py
aterzis-google: I was using beta because the paper calls it beta but I switched to weight_decay.

objax/optimizer/lars.py
aterzis-google: OK. I removed the LR decay logic and updated the doc string to indicate that the paper suggested calling the optimizer with a decaying LR.

objax/optimizer/lars.py
aterzis-google: Removed.

objax/optimizer/lars.py
aterzis-google: updated calculation of clipped_trust_ratio, now called rectified_trust_ratio, per offline discussion.

objax/optimizer/lars.py
```diff
@@ -0,0 +1,72 @@
...
+
+        for g, p, m in zip(grads, self.train_vars, self.m):
+            p_norm = jn.linalg.norm(p.value)
+            g_norm = jn.linalg.norm(g)
+            trust_ratio = self.tc * p_norm / (g_norm + self.weight_decay * p_norm + self.eps)
+            rectified_trust_ratio = jn.maximum(jn.logical_or(p_norm == 0, g_norm == 0), trust_ratio)
+            scaled_lr = lr * rectified_trust_ratio
+
+            g += self.weight_decay * p.value
+            m.value = self.momentum * m.value + scaled_lr * g
```
david-berthelot, line 71: A bit more concise:
```suggestion
            local_lr = lr * jn.maximum(jn.logical_or(p_norm == 0, g_norm == 0), trust_ratio)
            m.value = self.momentum * m.value + local_lr * (g + self.weight_decay * p.value)
```

david-berthelot (changes requested):

david-berthelot (approved):

PR #144, opened 2020-11-17
Add better error to objax.GradValues function for input_argnums

carlini:
Currently `input_argnums` expects a tuple. If you pass the integer 0, then the `or` treats this as if you passed nothing, and so  it passes an empty tuple and silently does the wrong thing. This fixes that to check the type.

### DIFF
--- a/objax/gradient.py
+++ b/objax/gradient.py
@@ -61,7 +61,7 @@ class GradValues(Module):
                 self.vc.assign(original_vc)
 
         assert isinstance(input_argnums, tuple) or input_argnums is None, \
-                f'Must pass a tuple of indices to input_argnums; received {input_argnums}.'
+            f'Must pass a tuple of indices to input_argnums; received {input_argnums}.'
         self.input_argnums = input_argnums or tuple()
 
         signature = inspect.signature(f)
--- a/objax/io/checkpoint.py
+++ b/objax/io/checkpoint.py
@@ -90,7 +90,7 @@ class Checkpoint:
             ckpt: full path to the restored checkpoint.
         """
         assert isinstance(vc, VarCollection), \
-            f"Must pass a VarCollection to save; received type {type(vc)}."
+            f'Must pass a VarCollection to restore; received type {type(vc)}.'
         if idx is None:
             all_ckpts = glob.glob(os.path.join(self.logdir, self.DIR_NAME, self.FILE_MATCH))
             if not all_ckpts:
--- a/objax/io/checkpoint.py
+++ b/objax/io/checkpoint.py
@@ -89,6 +89,8 @@ class Checkpoint:
             idx: index of the restored checkpoint.
             ckpt: full path to the restored checkpoint.
         """
+        assert isinstance(vc, VarCollection), \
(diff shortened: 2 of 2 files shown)


objax/gradient.py
```diff
@@ -60,7 +60,13 @@ def f_func(inputs_and_train_tensors: List[JaxArray],
             finally:
                 self.vc.assign(original_vc)
 
-        self.input_argnums = input_argnums or tuple()
+        if input_argnums is None:
+            self.input_argnums = tuple()
+        else:
+            assert isinstance(input_argnums, tuple), \
+                f"Must pass a tuple of indices to input_argnums; received {input_argnums}."
+            self.input_argnums = input_argnums
```
david-berthelot, line 68: I don't think we need an if statement, we can keep the code simple by changing the assert condition.
```suggestion
        assert isinstance(input_argnums, tuple) or input_argnums is None, \
                f'Must pass a tuple of indices to input_argnums; received {input_argnums}.'
        self.input_argnums = input_argnums or tuple()
```

Also nit, let's use single quotes to keep string consistency with rest of code.

david-berthelot (changes requested):

objax/gradient.py
```diff
@@ -60,7 +60,10 @@ def f_func(inputs_and_train_tensors: List[JaxArray],
             finally:
                 self.vc.assign(original_vc)
 
+        assert isinstance(input_argnums, tuple) or input_argnums is None, \
+                f'Must pass a tuple of indices to input_argnums; received {input_argnums}.'
```
david-berthelot, line 64: There's a flake8 error. Can you fix it?

david-berthelot (changes requested):

carlini:
Pushing fix

objax/io/checkpoint.py
```diff
@@ -89,6 +89,8 @@ def restore(self, vc: VarCollection, idx: Optional[int] = None):
             idx: index of the restored checkpoint.
             ckpt: full path to the restored checkpoint.
         """
+        assert isinstance(vc, VarCollection), \
+            f"Must pass a VarCollection to save; received type {type(vc)}."
```
david-berthelot, line 93: Nit: coding style consistency, also I believe you meant restore not save here.
```suggestion
            f'Must pass a VarCollection to restore; received type {type(vc)}.'
```

david-berthelot (changes requested):

objax/io/checkpoint.py
```diff
@@ -109,6 +111,8 @@ def save(self, vc: VarCollection, idx: int):
             vc: variables collection to save.
             idx: index of the new checkpoint where variables should be saved.
         """
+        assert isinstance(vc, VarCollection), \
+            f"Must pass a VarCollection to save; received type {type(vc)}."
```
david-berthelot, line 115:
```suggestion
            f'Must pass a VarCollection to save; received type {type(vc)}.'
```

david-berthelot (changes requested):

david-berthelot (approved):

PR #145, opened 2020-11-18
Give error if passing the wrong type (e.g., a module) to checkpoint save or restore

carlini:
Currently if someone accidentally calls save or restore on a module, things crash with an unhelpful error. This gives a better error.

### DIFF
--- a/objax/io/checkpoint.py
+++ b/objax/io/checkpoint.py
@@ -110,8 +110,7 @@ class Checkpoint:
             vc: variables collection to save.
             idx: index of the new checkpoint where variables should be saved.
         """
-        assert isinstance(vc, VarCollection), \
-            f"Must pass a VarCollection to save; received type {type(vc)}."
+        assert isinstance(vc, VarCollection), f'Must pass a VarCollection to save; received type {type(vc)}.'
         self.SAVE_FN(os.path.join(self.logdir, self.DIR_NAME, self.FILE_FORMAT % idx), vc)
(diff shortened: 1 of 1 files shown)


objax/io/checkpoint.py
```diff
@@ -89,6 +89,8 @@ def restore(self, vc: VarCollection, idx: Optional[int] = None):
             idx: index of the restored checkpoint.
             ckpt: full path to the restored checkpoint.
         """
+        assert isinstance(vc, VarCollection), \
+            f"Must pass a VarCollection to save; received type {type(vc)}."
```
david-berthelot, line 93:
```suggestion
        assert isinstance(vc, VarCollection), f'Must pass a VarCollection to restore; received type {type(vc)}.'
```

Nits: (1) use single quotes, (2) we allow 120 character lines in code and (3) I believe this function is for restore, the text was saying save.

objax/io/checkpoint.py
```diff
@@ -109,6 +111,8 @@ def save(self, vc: VarCollection, idx: int):
             vc: variables collection to save.
             idx: index of the new checkpoint where variables should be saved.
         """
+        assert isinstance(vc, VarCollection), \
+            f"Must pass a VarCollection to save; received type {type(vc)}."
```
david-berthelot, line 115:
```suggestion
        assert isinstance(vc, VarCollection), f'Must pass a VarCollection to save; received type {type(vc)}.'
```
Nits: (1) use single quotes and (2) we allow 120 character lines in code.

david-berthelot (changes requested):

david-berthelot (approved):

PR #152, opened 2020-11-20
Add jax.lax.scan to functional.

aterzis-google:


tests/scan.py
```diff
@@ -0,0 +1,31 @@
...
+
+import objax
+
+
+class TestScan(unittest.TestCase):
+    def test_scan(self):
+        def cell(carry, x):
+            return jn.array([2]) * carry * x, jn.array([2]) * carry * x
+
+        a, b = objax.functional.scan(cell, jn.zeros((2,)), jn.zeros((3,)))
```
david-berthelot, line 29: The test is a bit too light, like using zeros to test a multiplication leads to more zeros. Also you would not find out if the carry is properly as the first argument or not.
1. Use non-zero inputs
2. Make carry and output different values (right now they're the same)
3. Test return values (testing the values will test their shape as well).

david-berthelot (approved):

tests/scan.py
david-berthelot: Change requested

david-berthelot (changes requested):

tests/scan.py
aterzis-google: 1. Done
2. Done
2. Done

david-berthelot: I don't think 2 is what I had in mind, I wanted to return something different.

Right now it returns 
```
return jn.array([2]) * carry * x, jn.array([2]) * carry * x
```
Which are both the same, so it could return carry as output and the test wouldn't catch it.

What I in mind would be something like:
```
return jn.array([2]) * carry * x, jn.array([3]) * carry * x
```

aterzis-google: Thank you for clarifying and providing the suggestion. I've updated the code to do as suggested

david-berthelot (approved):

PR #153, opened 2020-11-20
Add support on objax.Parallel for list arguments.

jyh:
The default is True to preserve the behavior of splitting the batch
dimension across devices.

If False, arguments are passed directly to jax.pmap without reshaping.
This supports cases, for example, where an argument is a list of tensors
to be split across devices.

### DIFF
--- a/objax/module.py
+++ b/objax/module.py
@@ -227,8 +227,7 @@ class Parallel(Module):
                  vc: Optional[VarCollection] = None,
                  reduce: Callable[[JaxArray], JaxArray] = jn.concatenate,
                  axis_name: str = 'device',
-                 static_argnums: Optional[Tuple[int, ...]] = None,
-                 split_args: bool = True):
+                 static_argnums: Optional[Tuple[int, ...]] = None):
         """Parallel constructor.
 
         Args:
@@ -238,8 +237,6 @@ class Parallel(Module):
             axis_name: what name to give to the device dimension, used in conjunction with objax.functional.parallel.
             static_argnums: tuple of indexes of f's input arguments to treat as static (constants)).
                 A new graph is compiled for each different combination of values for such inputs.
-            split_args: if True, split non-static arguments along the batch dimension across devices;
-                otherwise, the first dimension of non-static arguments should equal the number of devices.
         """
         if not isinstance(f, Module):
             if vc is None:
@@ -262,7 +259,6 @@ class Parallel(Module):
         self.vc = vc or f.vars()
         self._call = jax.pmap(pmap, axis_name=axis_name, static_broadcasted_argnums=[x + 2 for x in static_argnums])
         self.__wrapped__ = f
-        self.split_args = split_args
 
     def device_reshape(self, x: JaxArray) -> JaxArray:
         """Utility to reshape an input array in order to broadcast to multiple devices."""
@@ -282,8 +278,7 @@ class Parallel(Module):
             f'Some variables were not replicated: {unreplicated}.' \
             'did you forget to call VarCollection.replicate on them?'
 
-        if self.split_args:
-            args = [x if i in self.static_argnums else self.device_reshape(x) for i, x in enumerate(args)]
+        args = [x if i in self.static_argnums else jax.tree_map(self.device_reshape, [x])[0] for i, x in enumerate(args)]
         output, changes = self._call(self.vc.tensors(), self.vc.subset(RandomState).tensors(), *args)
         self.vc.assign(changes)
         return jax.tree_map(self.reduce, output)
--- a/objax/module.py
+++ b/objax/module.py
@@ -278,7 +278,8 @@ class Parallel(Module):
             f'Some variables were not replicated: {unreplicated}.' \
             'did you forget to call VarCollection.replicate on them?'
 
-        args = [x if i in self.static_argnums else jax.tree_map(self.device_reshape, [x])[0] for i, x in enumerate(args)]
+        args = [x if i in self.static_argnums
+                else jax.tree_map(self.device_reshape, [x])[0] for i, x in enumerate(args)]
         output, changes = self._call(self.vc.tensors(), self.vc.subset(RandomState).tensors(), *args)
         self.vc.assign(changes)
         return jax.tree_map(self.reduce, output)
--- a/objax/module.py
+++ b/objax/module.py
@@ -227,7 +227,8 @@ class Parallel(Module):
                  vc: Optional[VarCollection] = None,
                  reduce: Callable[[JaxArray], JaxArray] = jn.concatenate,
                  axis_name: str = 'device',
-                 static_argnums: Optional[Tuple[int, ...]] = None):
+                 static_argnums: Optional[Tuple[int, ...]] = None,
+                 split_args: bool = True):
         """Parallel constructor.
 
         Args:
@@ -237,6 +238,8 @@ class Parallel(Module):
             axis_name: what name to give to the device dimension, used in conjunction with objax.functional.parallel.
             static_argnums: tuple of indexes of f's input arguments to treat as static (constants)).
                 A new graph is compiled for each different combination of values for such inputs.
+            split_args: if True, split non-static arguments along the batch dimension across devices;
+                otherwise, the first dimension of non-static arguments should equal the number of devices.
         """
         if not isinstance(f, Module):
             if vc is None:
@@ -259,6 +262,7 @@ class Parallel(Module):
         self.vc = vc or f.vars()
         self._call = jax.pmap(pmap, axis_name=axis_name, static_broadcasted_argnums=[x + 2 for x in static_argnums])
         self.__wrapped__ = f
+        self.split_args = split_args
 
     def device_reshape(self, x: JaxArray) -> JaxArray:
         """Utility to reshape an input array in order to broadcast to multiple devices."""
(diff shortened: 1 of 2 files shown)


objax/module.py
```diff
@@ -278,7 +282,8 @@ def __call__(self, *args):
             f'Some variables were not replicated: {unreplicated}.' \
             'did you forget to call VarCollection.replicate on them?'
 
-        args = [x if i in self.static_argnums else self.device_reshape(x) for i, x in enumerate(args)]
+        if self.split_args:
```
david-berthelot, line 285: After this we would either split everything or nothing. Do you envision use cases where one would want to only split some of the inputs? If so, should we instead make it a list of indexes to not split?

david-berthelot (changes requested):
Also would like Alex' opinion.

objax/module.py
jyh: I like that approach, too, to match static_argnums, happy to update.  Will wait on Alex.

One nit, I wonder why to do this reshaping at all, because it seems simpler to keep the arguments the same as pmap, in keeping with objax mission of simplicity and minimal layering, unlike e.g. Keras.  But anyway, this is just musing, not really important.

AlexeyKurakin: Actually, first I'll reply your question why we do reshaping. It's actually relevant to the review of the code

On the one hand this reshaping adds some additional complexity. On the other hand it allows us to pretty much use single-device and parallel code interchangeably with minimal modifications.
With the current approach you can pretty much take single-device code, replace `objax.Jit` with `objax.Parallel`, add call to `model.vars().replicate()` and it will be parallelized over multiple devices.

AlexeyKurakin: Now, let's talk about this change.

I think for each argument which is passed to objax.Parallel we probably want one of the following behaviors:
1) auto-reshape argument to split it across devices (current default behavior). One typical example - you pass your training examples that way, when you parallelizing single-device with minimal changes
2) static_argnums - arguments which are treated as compile-time constants during compilation. Right now they are broadcasted to all devices
3) *PROPOSED IN THIS PULL REQUEST* do no auto-reshape, instead just split argument across first axis.
4) *NEW ONE, NOT COVERED ANYWHERE*. broadcast argument to all devices. This is not the same as 2) because argument is not treated as compile time constant. One example of this use case - broadcasting current epoch or learning rate to all devices (example code: https://github.com/google/objax/blob/74e7739b0062977c2a78141f3a4e791cf6580902/examples/image_classification/imagenet_resnet50_train.py#L163 )

Current Objax code only implements 1) and 2). This pull request proposed 3), however there is also 4) with some valid usecases [which are handled by workaround]

So since we doing this change, we should think about it in the context of adding both 3) and 4)
Now, I'm not saying that 4) should be added in this pull request, we just need to think about design of this feature taking into account both 3) and 4)

david-berthelot: I agree 4 does not belong to this PR, so far this has not gotten in the way but definitely would be nice to have (I can some existing code becoming simpler thanks to it).

How the (3.b) I was discussing: can you envision some use cases where some of the arguments would be split and others would already be split. Or is it the case that all or nothing is good enough (and also it's simpler)?

Right now, my own inclination is the boolean (all or nothing) is good enough, but I'd like more feedback.

AlexeyKurakin: There are few ways to achieve this.

Two obvious ones:

A) We add something like `split_argnums` for 3) and `broadcast_argnums` for 4)  
i.e. if argument index is in `split_argnums` then do not reshape and instead split arguments across first axis.
if argument index is in `broadcast_argnums` then broadcast it per device

[we can come up with better names than `split_argnums`]

B1) provide a dictionary `args_sharding`. [or pick a better name]
Then the logic of how arguments are handled could be following:
```python
arg_idx = ... # argument index
if arg_idx not in args_sharding:
  # auto-reshape, i.e. current behavior
elif args_sharding[arg_idx] == SPLIT:
   # do not reshape, proposed behavior
elif args_sharding[arg_idx] == BROADCAST:
  # broadcast argument
```

B2) similar to B1, however `args_sharding` contains a **sharding function**
Arguments are handled in a following way
```
arg_value = ... # argument value
arg_idx = ... # argument index
sharding_fn = args_sharding.get(arg_idx, self.device_reshape)
arg_value = sharding_fn(arg_value)
```

Then if you need behavior 3) then `args_sharding[arg_idx] = lambda x: x`
If you need behavior 4) then `args_sharding[arg_idx] = lambda x: broadcast(x, jax.num_local_devices())`

Overall I'm leaning toward B2, but open to suggestions

AlexeyKurakin: Overall in the B2 case, code of Parallel will look like following:

```python
class Parallel(Module):

    def __init__(self,
                 f: Union[Module, Callable],
                 vc: Optional[VarCollection] = None,
                 reduce: Callable[[JaxArray], JaxArray] = jn.concatenate,
                 axis_name: str = 'device',
                 static_argnums: Optional[Tuple[int, ...]] = None,
                 args_sharding: Dict[int, Callable]):   # <-- new argument
        # init function ...
        # put args_sharding into self
        # and add handling of static_argnums
        self.args_sharding = dict(args_sharding)
        self.args_sharding.update({i: lambda x: x for i in static_argnums})

    def __call__(self, *args):
        unreplicated = ... 
        assert not unreplicated

        args = [self.args_sharding.get(i, self.device_reshape)(x) for i, x in enumerate(args)]
        output, changes = self._call(self.vc.tensors(), self.vc.subset(RandomState).tensors(), *args)
        self.vc.assign(changes)
        return jax.tree_map(self.reduce, output)
```

david-berthelot: Look quite elegant, the type signature for `Callable` would be `Callable[[JaxArray], jax.interpreters.pxla.ShardedDeviceArray]`, correct?

AlexeyKurakin: @david-berthelot 

Right now I don't see any obvious use case when you need to mix "auto-reshape" with "non auto-reshape".
But I would like to hear for specific use case for "non auto-reshape", maybe then it will be more clear when we may need to mix it.

I generally against just adding a bool flag for "non auto-reshape" because it generally feels like quick hack, which we can never properly fix in the future (without breaking user code) if we realize that it's not good enough.
Specifically, there are already two use cases (i.e.  3) and 4) ) when non-standard behavior is desired.
So let's say for this PR we add boolean flag, then for 4) we add yet another argument.
Later we figure out that we actually need to mix "non auto-reshape" with "auto-reshape"  and maybe add another type of behavior, so what do we do? keep adding arguments to Parallel sounds like an easy way to get something overcomplicated and hard to use.

That's why I'm suggesting to think about some generic enough solution which can accomodate both 3) and 4) without overbloating `Parallel`

david-berthelot: The thing is if a user is going to need complex logic to do their own sharding, then maybe it's better to let them do it before invoking a Parallel function and simply tell Parallel to treat its arguments as already sharded. So a bool would be sufficient in that context.

So the code would be exactly like original code (I just changed the argument name for clarification):
```
class Parallel(Module):
    def __init__(self,
                 ...
                 static_argnums: Optional[Tuple[int, ...]] = None,
                 args_already_sharded: bool = False):
        self.args_already_sharded = args_already_sharded

    def device_reshape(self, x: JaxArray) -> JaxArray:
        if not self.args_already_sharded:
            args = [x if i in self.static_argnums else self.device_reshape(x) for i, x in enumerate(args)]
```

For most easy cases, it's sufficient. For complex cases, just let the sharding to the user and set the flag `args_already_sharded=True`, it has the advantage to keep the `Parallel` code as simple as can be.

AlexeyKurakin: Technically speaking, even existing behavior sufficient for most cases (instead of passing `args_already_sharded` user may instead just reshape all arguments). So it's a question on how much convenience `Parallel` should provide to users.

Anyway, @david-berthelot how would you suggest handling broadcasting of arguments to parallel (i.e. broadcasting learning rate or epoch)? As we agreed implementation is out of scope of this pull request. Nevertheless I think we should keep in mind potential design of broadcasting while thinking about specific solution in this pull request

jyh: While I do like the flexibility of the args_sharding dict, it may be overkill.  The current auto-sharding on the batch dimension works perfectly for nearly everybody.  Then there is the special-case user like me who is splitting across a non-batch dimension (for example, I am mapping the 6 faces of a cube onto 6 out of 8 devices, and I use 5D tensors FBCHW, where F is cube face).  For the non-batch case, the auto-sharding doesn't make sense for any of the arguments.

The main technical issue now is that the auto-sharder disables pmap support for List[JaxArray] arguments, where the list elements are to be sharded by pmap.

A more obscure, but workable, solution is to turn off auto-sharding for List arguments.

`args = [x if i in self.static_argnums or type(x)==list else self.device_reshape(x) for i, x in enumerate(args)]`

We wouldn't need to change the module arguments, giving more flexibility to decide in the future how to do broadcast arguments (4).

david-berthelot: @AlexeyKurakin So far I've handled broadcasting before calling Parallel, for example this code does that for passing the training step to a Parallel instance: https://github.com/google/objax/blob/master/examples/fixmatch/libml/train.py#L98

The functionality is simple, we could have a broadcast to device function for it if it makes life easier for users and we'd put it in the same place as map to device.

@jyh Did we support lists in the first place in `objax.Parallel`? I don't think so and I don't think we wanted to. I'm not sure adding list support aligns with my ideals, I'd need to understand better what your use case is.

jyh: @david-berthelot No, objax.Parallel does not support lists, but jax.pmap does.  I want to support a case similar to what objax is doing with m.vars().replicate(), which produces a List[JaxArray] that pmap splits across devices, where the JaxArray have varying shapes (this feature is used in the implementation on objax.Parallel).  In my case, the thing I am replicating is the description of how the devices are stitched together, so that I can implement a global Conv2D spanning devices, using lax.ppermute to copy borders between devices.

The reason I use jax is so I can do funky stuff like this, it is very cool vs. TF or Keras, for example.  But objax is nicer to program in than pure jax, and it has been really great, except this one small feature of jax.pmap was disabled.  

From some idealistic POV, we want to keep as much as possible of that jax generality and simplicity that makes it so attractive.  But I will not stand my ground on this:). Maybe there is some other workaround I can try.

david-berthelot (approved):
Note: the author did reshaping before parallel, which solved the issue while keeping the API simple.

PR #161, opened 2020-11-21
Update the instructions about how to download vgg19.npy.

aterzis-google:


david-berthelot (commented):
Do these instructions work with `pip install`?
I know they work when you clone github for development, but I don't know in user install scenario.

PR #171, opened 2020-12-11
Adding support for __repr__, fixes #142

### LINKED ISSUE #142: Add __repr__ method for modules
To help inspect a model's architecture.

david-berthelot:
See sample below:
```python
import functools

import objax

m = objax.nn.Sequential([
    objax.nn.Linear(2, 3),
    objax.functional.leaky_relu,
    objax.nn.Sequential([
        functools.partial(objax.functional.leaky_relu, negative_slope=0.5),
        objax.nn.Linear(4, 5)]),
    objax.functional.tanh,
    objax.nn.MovingAverage((2, 3), 4, 0.2),
    objax.nn.ExponentialMovingAverage((5, 6)),
    objax.nn.BatchNorm((3, 4, 5), (1, 7)),
    objax.nn.BatchNorm0D(8),
    objax.nn.BatchNorm1D(9),
    objax.nn.BatchNorm2D(19),
    objax.nn.Conv2D(4, 5, k=3),
    objax.nn.ConvTranspose2D(4, 5, k=3),
    objax.nn.Dropout(0.7),
    objax.nn.SyncedBatchNorm((3, 4, 5), (1, 7)),
    objax.nn.SyncedBatchNorm0D(8),
    objax.nn.SyncedBatchNorm1D(9),
    objax.nn.SyncedBatchNorm2D(19),
])
print(m)
```

Output:
```plain
objax.nn.Sequential(
  [0] objax.nn.Linear(nin=2, nout=3, use_bias=True, w_init=xavier_normal(*, gain=1))
  [1] leaky_relu(*, negative_slope=0.01)
  [2] objax.nn.Sequential(
    [0] leaky_relu(*, negative_slope=0.5)
    [1] objax.nn.Linear(nin=4, nout=5, use_bias=True, w_init=xavier_normal(*, gain=1))
  )
  [3] tanh
  [4] objax.nn.MovingAverage(shape=(2, 3), buffer_size=4, init_value=0.2)
  [5] objax.nn.ExponentialMovingAverage(shape=(5, 6), momentum=0.999, init_value=0)
  [6] objax.nn.BatchNorm(dims=(3, 4, 5), redux=(1, 7), momentum=0.999, eps=1e-06)
  [7] objax.nn.BatchNorm0D(nin=8, momentum=0.999, eps=1e-06)
  [8] objax.nn.BatchNorm1D(nin=9, momentum=0.999, eps=1e-06)
  [9] objax.nn.BatchNorm2D(nin=19, momentum=0.999, eps=1e-06)
  [10] objax.nn.Conv2D(nin=4, nout=5, k=(3, 3), strides=(1, 1), dilations=(1, 1), groups=1, padding='SAME', use_bias=True, w_init=kaiming_normal(*, gain=1))
  [11] objax.nn.ConvTranspose2D(nin=5, nout=4, k=(3, 3), strides=(1, 1), dilations=(1, 1), padding='SAME', use_bias=True, w_init=kaiming_normal(*, gain=1))
  [12] objax.nn.Dropout(keep=0.7)
  [13] objax.nn.SyncedBatchNorm(dims=(3, 4, 5), redux=(1, 7), momentum=0.999, eps=1e-06)
  [14] objax.nn.SyncedBatchNorm0D(nin=8, momentum=0.999, eps=1e-06)
  [15] objax.nn.SyncedBatchNorm1D(nin=9, momentum=0.999, eps=1e-06)
  [16] objax.nn.SyncedBatchNorm2D(nin=19, momentum=0.999, eps=1e-06)
)
```

### DIFF
--- a/objax/module.py
+++ b/objax/module.py
@@ -22,7 +22,7 @@ import jax.numpy as jn
 from jax.interpreters.pxla import ShardedDeviceArray
 
 from objax.typing import JaxArray
-from objax.util import override_args_kwargs, positional_args_names
+from objax.util import class_name, override_args_kwargs, positional_args_names, repr_function
 from objax.variable import BaseVar, RandomState, VarCollection
 
 
@@ -115,6 +115,10 @@ class ForceArgs(Module):
         args, kwargs = override_args_kwargs(self.__wrapped__, args, kwargs, self.forced_kwargs)
         return self.__wrapped__(*args, **kwargs)
 
+    def __repr__(self):
+        args = ', '.join(f'{k}={repr(v)}' for k, v in self.forced_kwargs.items())
+        return f'{class_name(self)}(module={repr_function(self.__wrapped__)}, {args})'
+
 
 class ModuleList(Module, list):
     """This is a replacement for Python's list that provides a vars() method to return all the variables that it
@@ -142,6 +146,17 @@ class ModuleList(Module, list):
         if isinstance(key, slice):
             return ModuleList(value)
         return value
+
+    def __repr__(self):
+        def f(x):
+            if not isinstance(x, Module) and callable(x):
+                return repr_function(x)
+            x = repr(x).split('\n')
+            x = [x[0]] + ['  ' + y for y in x[1:]]
+            return '\n'.join(x)
+
+        entries = '\n'.join(f'  [{i}] {f(x)}' for i, x in enumerate(self))
+        return f'{class_name(self)}(\n{entries}\n)'
 
 
 class Function(Module):
@@ -179,6 +194,9 @@ class Function(Module):
 
         return from_function
 
+    def __repr__(self):
+        return f'{class_name(self)}(f={repr_function(self.__wrapped__)})'
+
 
 class Jit(Module):
     """JIT (Just-In-Time) module takes a function or a module and compiles it for faster execution."""
@@ -195,6 +213,7 @@ class Jit(Module):
             static_argnums: tuple of indexes of f's input arguments to treat as static (constants)).
                 A new graph is compiled for each different combination of values for such inputs.
         """
+        self.static_argnums = static_argnums
         if not isinstance(f, Module):
             if vc is None:
                 raise ValueError('You must supply the VarCollection used by the function f.')
@@ -217,6 +236,9 @@ class Jit(Module):
         output, changes = self._call(self.vc.tensors(), kwargs, *args)
         self.vc.assign(changes)
         return output
+
+    def __repr__(self):
+        return f'{class_name(self)}(f={self.__wrapped__}, static_argnums={self.static_argnums or None})'
 
 
 class Parallel(Module):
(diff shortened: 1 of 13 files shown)


objax/nn/layers.py
```diff
@@ -187,6 +203,13 @@ def __call__(self, x: JaxArray) -> JaxArray:
             y += self.b.value
         return y
 
+    def __repr__(self):
+        args = dict(nin=self.w.value.shape[2] * self.groups, nout=self.w.value.shape[3], k=self.w.value.shape[:2],
```
AlexeyKurakin, line 207: use `self.nin` instead of `self.w.value.shape[2] * self.groups`?
Or maybe remove `self.nin` completely if it's actually not used anywhere

objax/nn/layers.py
```diff
@@ -236,6 +259,13 @@ def __call__(self, x: JaxArray) -> JaxArray:
...
```
AlexeyKurakin, line 263: In the init function, when constructor of superclass is called `nin` and `nout` are flipped for `ConvTranspose2D` compared to superclass

objax/module.py
```diff
@@ -143,6 +147,17 @@ def __getitem__(self, key: Union[int, slice]):
             return ModuleList(value)
         return value
 
+    def __repr__(self):
+        def f(x):
+            if not isinstance(x, Module) and callable(x):
+                return repr_function(x)
+            x = repr(x).split('\n')
+            x = [x[0]] + ['  ' + y for y in x[1:]]
```
AlexeyKurakin, line 155: Will this work correctly for multiple levels of nesting of Sequential?
It looks like it only adds two spaces in front of lines here and then concatenates everything using newline.

Could you add test of with several level of depth of Sequential to make sure that it works, i.e.:
```
m = objax.Sequential(
   op11,
   objax.Sequential(
     op21,
     objax.Sequential(
       op31,
       op32,
       objax.Sequential(op41, op42),
     )
   )
   op12
)
```

AlexeyKurakin (commented):
Two high level comments:

1) In several places in the code you do something like following:
```
        args = dict(...)
        args = ', '.join(f'{x}={y}' for x, y in args.items())
```

As far as I understand, order of elements in Python dictionary is not guaranteed and may change (from run to run). Thus `repr` will also change.
I don't know if there are reasons to ensure that `repr` is fixed for the same object, but maybe worth thinking.

If you want fixed order, then maybe can use `OrderedDict` or just explicitly specify order


2) base class `objax.Module` does not provide implementation of `__repr__` thus all non-standard modules won't have a repr. This includes some modules from objax zoo as well as all user modules.
Is it by design? or should we add some default repr to objax.Module which lists all submodules?

objax/module.py
david-berthelot: Yes it does, there's a unit test for it.

david-berthelot:
General comments:
1. (dict order): https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6
2. No default `Module.__repr__`: it's by design, the `__repr__` represents the arguments used to create the object (not the fields it contains). I don't know of an automatic way to do it.

objax/nn/layers.py
david-berthelot: Removed `self.nin` to avoid incoherencies.

AlexeyKurakin (approved):

Issue #109, opened 2020-10-15
Explore Objax to Tensorflow conversion


AlexeyKurakin:
Some other JAX frameworks provide API to convert JAX models into Tensorflow:

1. https://trax-ml.readthedocs.io/en/latest/notebooks/tf_numpy_and_keras.html#2.-Convert-Trax-to-Keras
2. https://source.corp.google.com/piper///depot/google3/third_party/py/jax/experimental/jax2tf/examples/stax_to_tf_module.py
3. https://github.com/google/jax/tree/master/jax/experimental/jax2tf

Such conversion might be useful because Tensorflow allow to save trained models in SavedModel format (which contains both weights and network architecture) to be later used in production settings.

AlexeyKurakin:
I have run some preliminary experiments on conversion using jax2tf. It semi-works, however in certain cases hits JAX error.

Following example works:
```python
wrn_width = 2            # Width of WideResNet
wrn_depth = 28           # Depth of WideResNet
batch_size = 4

# Model
model = WideResNet(nin=3, nclass=10, depth=wrn_depth, width=wrn_width)
model_vars = model.vars()

# Prediction operation
predict_op = lambda x: objax.functional.softmax(model(x, training=False))
predict_op = objax.Jit(predict_op, model_vars)

# Run prediction on random batch
x = objax.random.normal((batch_size, 3, 32, 32))
pred_y = predict_op(x)
print(pred_y)

# Convert model to Tensorflow and run it on the same batch
predict_tf = jax2tf.convert(predict_op)
print(predict_tf(np.array(x)))
```

However, attempt to re-run `predict_tf` one more time causes `UnexpectedTracerError: Encountered an unexpected tracer. Perhaps this tracer escaped through global state from a previously traced function.`. Which is probably related to the issue we observing in https://github.com/google/objax/issues/158

david-berthelot:
I don't see how it could escape though, it looks like something specific to jax2tf somehow.
Otherwise the error would also show up when running `predict_op` twice (the JAX jitted one).

AlexeyKurakin:
I guess you're right. Maybe I have to add functional wrapper for jax2tf (similar to Jit, Grad, etc...). I'll look into this more

AlexeyKurakin:
After further investigation I made a version which seems to be working without issues:

```python
class Objax2Tf(tf.Module):

  def __init__(self, module: objax.Module):
[...]

Issue #158, opened 2020-11-21
Calling objax.Grad twice on a model with BatchNorm with respect to input params crashes


carlini:
The following code crashes.

```python
    import objax
    import jax.numpy as jn
    import numpy as np

    mod = objax.nn.Sequential([objax.nn.Conv2D(2, 4, 3),
                               objax.nn.BatchNorm2D(4)])

    def ell(x):
        return jn.sum(mod(x, training=True))

    m1 = objax.Grad(ell, {}, (0,))
    print(m1(np.ones((8*8*8,2,10,10))))

    m2 = objax.Grad(ell, {}, (0,))
    print(m2(np.ones((8*8*8,2,10,10)))) # boom!
```

The reason this happens is because the first time we call `grad()`, jax traces the function with a tracer. We update, AND SAVE, the batch norm state as the tracer object. Then, the second time we trace, we find the old tracer and blow up.

This should be fixed by making sure that we never save a jax tracer object into the BatchNorm state.

AlexeyKurakin:
We have to save (temporary) jax tracer into StateVars for JAX machinery to work correctly. However it has to be "restored" later.

So this gets to another broader issue of automatically figuring out list of variables used by a function.
If we know a list of variables then we can properly handle this situation as well as some other (like auto-detect variables needed for JIT)

Thus the next step would be to figure out the way to track all Objax variables used by a function.
One possible way is to make our own Objax Tracer object to trace variables used by functions.

Next step is to explore and prototype how to make such tracer work with BatchNorm.

AlexeyKurakin:
Here is a code snipped which encapsulates this example into unit test: https://github.com/google/objax/pull/189/files

This code snippet also includes fix "don't save tracer into stateVars". Unfortunately this is not correct fix and it lead to multiple failures in many unit tests.

The real problem here is not the fact that Tracers are saved into variables. Temporary saving tracers is actually necessary for Objax to work. The problem is that after tracing is done, variable values should be restored back to what is was before tracing.
To do this, `Grad` (or any other Objax transformation) has to know what variables were used inside the function which was called.

PR #176, opened 2020-12-15
Adding Objax2Tf converter

AlexeyKurakin:
Issue #109


objax/util/objax2tf.py
```diff
@@ -0,0 +1,73 @@
...
+            original_values = module_vars.tensors()
+            try:
+                module_vars.assign(tensor_list)
+                return module(*args, **kwargs)
+            finally:
+                module_vars.assign(original_values)
+
+        tf_function = jax2tf.convert(wrapped_op)
+        self._all_vars = [tf.Variable(v) for v in module_vars.tensors()]
+        self._call = tf_function
```
david-berthelot, line 68:
```suggestion
        self.tf_vars = [tf.Variable(v) for v in module_vars.tensors()]
        self.tf_call = tf_function
```

objax/util/objax2tf.py
```diff
@@ -0,0 +1,73 @@
...
+
+    @tf.function(autograph=False)
+    def __call__(self, *args, **kwargs):
+        """Calls Tensorflow function which was generated from Objax module."""
+        return self._call(self._all_vars, kwargs, *args)
```
david-berthelot, line 73:
```suggestion
        return self.tf_call(self.tf_vars, kwargs, *args)
```
Generally speaking, I prefer non private variables unless we really want to explicitly forbid their access (like `_value` in `objax.TrainVar`).

tests/objax2tf.py
```diff
@@ -0,0 +1,87 @@
...
+        model = WideResNet(NCHANNELS, NCLASSES, depth=4, width=1)
+        model_vars = model.vars()
+        # Prediction op without JIT
+        predict_op = lambda x: objax.functional.softmax(model(x, training=False))
+        predict_tf = objax.util.Objax2Tf(objax.Function(predict_op, model_vars))
+        # Compare results
+        self.verify_converted_predict_op(predict_op, predict_tf,
+                                         shape=(BATCH_SIZE, NCHANNELS, IMAGE_SIZE, IMAGE_SIZE))
+        # Predict op with JIT
+        predict_op_jit = objax.Jit(predict_op, model_vars)
```
david-berthelot, line 54:
```suggestion
        predict_op = objax.nn.Sequential([objax.ForceArgs(model, training=False), objax.functional.softmax])
        predict_tf = objax.util.Objax2Tf(predict_op)
        # Compare results
        self.verify_converted_predict_op(predict_op, predict_tf,
                                         shape=(BATCH_SIZE, NCHANNELS, IMAGE_SIZE, IMAGE_SIZE))
        # Predict op with JIT
        predict_op_jit = objax.Jit(predict_op)
```

tests/objax2tf.py
```diff
@@ -0,0 +1,87 @@
...
+        # Compare results
+        self.verify_converted_predict_op(predict_op_jit, predict_tf_jit,
+                                         shape=(BATCH_SIZE, NCHANNELS, IMAGE_SIZE, IMAGE_SIZE))
+
+    def test_savedmodel_wrn(self):
+        model_dir = tempfile.mkdtemp()
+        # Make a model and convert it to TF
+        model = WideResNet(NCHANNELS, NCLASSES, depth=4, width=1)
+        model_vars = model.vars()
+        predict_op = objax.Jit(lambda x: objax.functional.softmax(model(x, training=False)), model_vars)
```
david-berthelot, line 65:
```suggestion
        predict_op = objax.Jit(objax.nn.Sequential([objax.ForceArgs(model, training=False), objax.functional.softmax]))
```

david-berthelot (changes requested):
Proposed some cosmetic changes.

objax/util/objax2tf.py
AlexeyKurakin: `Objax2Tf` is a subclass of TF Module. `tf.Module` already provides `.variable` public property which returns all variables. Thus there is no need to expose it again.
Also `call` should be private because it's intended for very specific way of calling underlying function, which generally is not needed and should not be used outside of class (similar to `Jit._call` or `Parallel._call`).

If it adds clarity I can rename these two into `self._tf_vars` and `self._tf_call`

objax/util/objax2tf.py
AlexeyKurakin: See my previous comment regarding private/public

tests/objax2tf.py
AlexeyKurakin: What is the reason for this change?

tests/objax2tf.py
AlexeyKurakin: Same as above - what is the reason for this change?

AlexeyKurakin:
Please take another look

objax/util/objax2tf.py
david-berthelot: Thanks for the clarification, sure let's rename with `_tf_vars` and `_tf_call`

tests/objax2tf.py
david-berthelot: I'm trying to move away from functions as arguments: https://github.com/google/objax/issues/168
Plus it's the features of Objax more efficiently, and no need to pass vars to jit etc...

tests/objax2tf.py
david-berthelot: Moving away from functions since it can lead to error prone behavior (like not passing vars to Grad which led to errors like #158 ).

AlexeyKurakin:
@david-berthelot  please take another look

objax/util/objax2tf.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        assert hasattr(tf, '__version__'), 'Tensorflow must be installed for Objax2Tf to work.'
+        assert tf.__version__ >= '2.0', 'Objax2Tf works only with Tensorflow 2.'
+        assert isinstance(module, Module), 'Input argument to Objax2Tf must be an Objax module.'
+
+        super().__init__()
+
+        module_vars = module.vars()
+
+        def wrapped_op(tensor_list: List[JaxArray], kwargs, *args):
```
david-berthelot, line 58: Just putting a note here for us to think about (I'm going to merge regardless for now).

Say we do:
```python
import objax

m = objax.nn.BatchNorm0D(3)
tfm = objax.util.Objax2Tf(m)

x = objax.random.uniform((10, 3))
# What do we expect with? E.g. should batch statistics be updated (running_mean, running_var)? 
y = tfm(x, training=True)
```

david-berthelot (approved):

PR #178, opened 2020-12-15
Implementing 2 phases DP-SGD

lberrada:
This PR implements a two-phase algorithm for per-sample gradient clipping with the goal of improving memory efficiency for the training of private deep models. The two steps are: (1) accumulate the norms of the gradient per sample and (2) use those norm values to perform a weighted backward pass that is equivalent to per-sample clipping. The user can choose whether to use this new algorithm or the currently implemented one through a boolean argument.

The unit-tests have been adapted to check results for both algorithms.

Let me know if this fits well!

### DIFF
--- a/objax/privacy/dpsgd/gradient.py
+++ b/objax/privacy/dpsgd/gradient.py
@@ -39,7 +39,8 @@ class PrivateGradValues(Module):
                  l2_norm_clip: float,
                  microbatch: int,
                  batch_axis: Tuple[Optional[int], ...] = (0,),
-                 keygen: random.Generator = random.DEFAULT_GENERATOR):
+                 keygen: random.Generator = random.DEFAULT_GENERATOR,
+                 use_norm_accumulation: bool = False):
         """Constructs a PrivateGradValues instance.
 
         Args:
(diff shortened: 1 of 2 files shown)


shs037 (commented):
Thank you very much!
Two minor comments:
1. We changed the API to favor the usage of `@Function.with_vars(gv.vars())` in replace of specifying the variables (e.g. `gv.vars()`) in `Vectorize` explicitly (4e925a367e6de7e86d1fb4cd782a000b6199865c). Could you make the change accordingly?
2. There is another minor change that allows `GradValues` to return more than one output (7f6f2f7721e144ed8c2db2b45781e7c84a05076d). You may check if that affect anything.

Another question: would you mind writing a few sentences on when you would suggest users to use your algorithm e.g. always try it / try it when batch size is large / try it when running a certain kind of NN? 
(No need to be very rigorous, but just something users can refer to when doing dpsgd training.)

lberrada:
Hi, sorry for the late reply, I was on vacation.

I have tried to address all comments, let me know if this is fine or if there is anything else to change.

NB: I have squashed all commits into a single one.

shs037 (approved):

lberrada:
@shs037 thanks for the review! It now looks like a test is failing in a part of the code that is independent of this PR? Let me know if there is something that I can do on my end

shs037:
Sorry we're trying to fix it and will let you know asap.

aterzis-google:
@lberrada please sync with master and this will solve the problem with failing tests

lberrada:
Done, thanks!

PR #183, opened 2020-12-16
More numerically stable batch norm. Fixes #182

### LINKED ISSUE #182: Improve BatchNorm numerical stability
Instead of computing variance as E[x^2] - E[x]^2, use E[(x - E[x])^2]

david-berthelot:


tests/loss.py
```diff
@@ -62,7 +62,7 @@ def test_on_mean_squared_log_error(self):
         y = jn.array([[2.5, 5, 4, 8]], dtype=jn.float32)
         output = objax.functional.loss.mean_squared_log_error(x, y, keep_axis=(0,))
         expected = jn.array([0.039730113], dtype=jn.float32)
-        self.assertAlmostEqual(jn.abs(output - expected).sum(), 0, delta=1e-12)
+        self.assertAlmostEqual(jn.abs(output - expected).sum(), 0, delta=1e-7)
```
AlexeyKurakin, line 65: `expected` is a hardcoded number.
So maybe instead of changing the `delta`, value of `expected` should be updated

AlexeyKurakin (approved):

Issue #197, opened 2021-01-29
Prototype automatic tracing of variables used by a function


AlexeyKurakin:
Typically user has to specify list of variables used by a function:

```python
model = WideResNet(nin=3, nclass=10, depth=wrn_depth, width=wrn_width)

@objax.Function.with_vars(model.vars())
def loss_fn(x, label):
  logit = model(x, training=True)
  return objax.functional.loss.cross_entropy_logits_sparse(
      logit, label).mean()
```

The goal of the tracing is to automatically detect used variables, so the code could look like following:
```python
model = WideResNet(nin=3, nclass=10, depth=wrn_depth, width=wrn_width)

@objax.Function.auto_vars
def loss_fn(x, label):
  logit = model(x, training=True)
  return objax.functional.loss.cross_entropy_logits_sparse(
      logit, label).mean()
```

PR #198, opened 2021-01-29
Initial implementation of variable tracer wit AST tree

AlexeyKurakin:
Implementation of https://github.com/google/objax/issues/197

### DIFF
--- a/objax/__init__.py
+++ b/objax/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2020 Google LLC
+# Copyright 2021 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -30,6 +30,7 @@ from . import util
 from ._version import __version__
 from .constants import *
 from .gradient import *
+from .module_base import *
 from .module import *
 from .variable import *
 
--- a/objax/gradient.py
+++ b/objax/gradient.py
@@ -1,4 +1,4 @@
-# Copyright 2020 Google LLC
+# Copyright 2021 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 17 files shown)


david-berthelot (commented):
Generally speaking the code is amazingly elegant and concise.

- The AST part I can't say I understand mostly because I never looked into the Transformer. But it looks very readable and debuggable.
- One question though: why did you move Module to a new file and is it really necessary?

AlexeyKurakin:
@david-berthelot 
I moved Module to separate file to avoid circular dependency which happens because I add `Function.auto_vars` method.
In this case `objax/util/tracing.py` depends on `Module`; `Function` depends on `objax/util/tracing.py` and both `Module` and `Function` are located within the same file.

I though about following ways to deal with circular dependency:

1) Don't add `Function.auto_vars` static method.
In this case user has to do following:
```python
def loss(...):
    # loss function

loss = Function(loss, objax.util.tracing.find_user_variables(loss))
```
This might be fine for now while we experimenting with the feature. But eventually this is not very convenient syntax for automatic variable detection.

2) Move `objax.Module` into separate file `module_base.py` (or we can change name to something else). Then `module_base.py` won't depend on anything else, thus circular dependency would be eliminated.
This is what I implemeneted in the pull request

3) Move `objax.Function` into separate file. For example, we can call it `function.py`. In this case `objax/util/tracing.py` will depends on `module.py` and `module.py` won't depend on `tracing.py`. So circular dependency would be eliminated.

4) Define `Function.auto_vars()` inside `tracing.py`. This is somewhat more complicated solution and I'm not entirely sure whether it would work or not.
It could look like following:
```python
# in module.py

class Function(Module):

    @staticmethod
    def auto_vars(f: Callable):
        """Turns a function into module by automatically detecting used Objax variables. Could be used as a decorator."""
        # This function is implemented in tracing.py to avoid circular dependency
        raise NotImplementedError()

# in objax/util/tracing.py

from objax.module import Function

def _auto_vars(f: Callable):
    return Function(f, find_used_variables(f))

# I don't know exact syntax, but idea is following:
set_attr(Function, 'auto_vars', _auto_vars)
```



Among all of this solutions, I think we can do \#1 or some variation of \#1 temporarily, but eventually we better add `Function.auto_vars` static method. Solution \#4 sounds may work, but it looks unnecessarily confusing to me.
So for now I picked \#2, but I don't have strong preference.

Let me know if you have other ideas how to deal with it.

AlexeyKurakin:
I addressed issue with moving module into separate file and created a new pull request https://github.com/google/objax/pull/205 which is an implementation of tracing ready for review.
Closing this pull request, since new one supersedes it.

PR #202, opened 2021-02-06
Learning rate scheduler

kihyuks:
Learning rate scheduler


objax/scheduler/scheduler.py
```diff
@@ -0,0 +1,126 @@
...
+__all__ = ['LinearAnnealing', 'StepDecay']
+
+
+from typing import List, Tuple, Union
+
+import jax.numpy as jn
+
+
+class Scheduler:
+    def __init__(self,
```
AlexeyKurakin, line 24: Some high level questions/suggestions about scheduler:

1. I think we discussed in the past that, maybe we want to use floating point `step` or `progress` counter instead of integer step counter. This way user may choose whether to use it as actual step counter, as fractional epoch or whatever. Right now it looks like you using integer step counter everywhere in the code
2. Currently the base class takes three arguments `is_cycle`, `max_step` and `base_lr` to the `__init__` function. However it looks to me that only `base_lr` is generic enough that almost every LR scheduler have it, while `is_cycle` and `max_step` sounds too specific and only used by some schedulers but not others.
3. It seems like scheduler tries to automatically increment step number. I think it's generally bad idea and may lead to confusion and bugs, especially given how Objax tracks training steps in other examples. In particular, you can not call scheduler twice at the same training step, otherwise step counter would increase. On the other hand calling scheduler twice may be convenient when you want to save current LR into summary or print it for debugging

AlexeyKurakin (changes requested):

objax/scheduler/scheduler.py
```diff
@@ -0,0 +1,126 @@
+# Copyright 2020 Google LLC
```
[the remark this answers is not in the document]
kihyuks: Will follow option 1.

objax/scheduler/scheduler.py
kihyuks: Done

objax/scheduler/scheduler.py
```diff
@@ -0,0 +1,126 @@
...
+            base_lr: base learning rate.
+            is_cycle: trigger cyclical learning rate multiplier when step
+                exceeds max_step.
+        """
+        self.step = 0
+        self.max_step = max_step
+        self.base_lr = base_lr
+        self.is_cycle = is_cycle
+
+    def multiplier(self, step):
```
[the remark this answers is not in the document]
kihyuks: Done.

kihyuks:
@AlexeyKurakin would you please take another look? thanks

objax/optimizer/scheduler.py
```diff
@@ -0,0 +1,112 @@
...
+        return 1.0 - (step / (self.max_step - 1)) * (
+            1.0 - self.min_lr_multiplier)
+
+
+class StepDecay(Scheduler):
+    def __init__(self,
+                 step_size: Union[float, List, Tuple],
+                 base_lr: float = 1.0,
+                 gamma: float = 0.1,
+                 **kwargs):
```
AlexeyKurakin, line 92: why do you need `**kwargs` here?
`LinearAnnealing` does not have it

objax/optimizer/scheduler.py
```diff
@@ -0,0 +1,112 @@
...
+    def multiplier(self, step: float):
+        """Returns linear annealing learning rate multiplier."""
+
+        # If is_cycle, we use the remainder of step; otherwise, we stop update.
+        if self.is_cycle:
+            step = jn.remainder(step, self.max_step)
+        else:
+            step = jn.minimum(step, self.max_step)
+
+        return 1.0 - (step / (self.max_step - 1)) * (
```
AlexeyKurakin, line 83: if we allow steps to be floating point number then this should be `step / self.max_step` instead of `step / (self.max_step - 1)`

kihyuks: Done.

objax/optimizer/scheduler.py
kihyuks: Removed kwargs.

objax/optimizer/__init__.py
```diff
@@ -17,3 +17,4 @@
 from .lars import *
 from .momentum import *
 from .sgd import *
+from .scheduler import *
```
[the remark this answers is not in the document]
kihyuks: Done.

kihyuks:
Thanks @AlexeyKurakin for comments. I pushed the code after modification.

AlexeyKurakin (approved):

Issue #111, opened 2020-10-16
Remove need for .value when referring to internal param values.


peterjliu:
When referring to a Module's internal variables in __call__, one needs to use `self.x.value` instead of simply `self.x`. It'd be nice to enable this syntactic sugar to improve the readability of complex math expressions. For example, `tf.module` allows this

 class Dense(tf.Module):
   def __init__(self, in_features, out_features, name=None):
     super(Dense, self).__init__(name=name)
     self.w = tf.Variable(
       tf.random.normal([in_features, out_features]), name='w')
     self.b = tf.Variable(tf.zeros([out_features]), name='b')
   def __call__(self, x):
     y = tf.matmul(x, self.w) + self.b
     return tf.nn.relu(y)

AlexeyKurakin:
I will implement it after JAX team will make this change: https://github.com/google/jax/pull/4725

PR #207, opened 2021-03-13
Objax variables could be used in place of native JAX arrays

AlexeyKurakin:
This is a prototype implementation of https://github.com/google/objax/issues/111

This is reasonably well tested and complete implementation of JAX duck typing, which allows to use Objax variables in mathematical expressions without explicitly calling `.value`.
This change does not include documentation, updates of examples, etc... which will be done in future pull requests.

### DIFF
--- a/objax/io/ops.py
+++ b/objax/io/ops.py
@@ -49,14 +49,17 @@ def load_var_collection(file: Union[str, IO[BinaryIO]],
         file = open(file, 'rb')
     data = np.load(file, allow_pickle=False)
     name_index = {renamer(k): str(i) for i, k in enumerate(data['names'])}
-    name_vars = collections.defaultdict(list)
+    var_names = collections.defaultdict(list)
+    var_values = {}
     for k, v in vc.items():
         if isinstance(v, TrainRef):
             v = v.ref
-        name_vars[v].append(k)
+        var_names[id(v)].append(k)
+        var_values[id(v)] = v
     misses = []
     used_vars = set()
-    for v, names in name_vars.items():
+    for var_id, names in var_names.items():
+        v = var_values[var_id]
         for name in names:
             index = name_index.get(name)
             if index is not None:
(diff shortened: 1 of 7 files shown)


AlexeyKurakin:
One drawback of the proposed design is that it requires to define a bunch of `__OP__` methods on `BaseVar`.
This is a python limitation, it does not have any elegant workaround and it's needed to support things like `var1 + var2` where `var1` and `var2` are Objax variables.
It somewhat bloats the `BaseVar` class.

While there is no way to avoid it, it is possible to "move" definitions of `__OP__` methods outside of the `BaseVar` class. For example, JAX does it this way: https://github.com/google/jax/blob/80966fe5bfb123a9b7b0200eb9aa53bcef5f1a2b/jax/_src/numpy/lax_numpy.py#L5390

In case of `BaseVar` it could be done in a following way:


```python
# Definition of BaseVar
class BaseVar(abc.ABC):
    def __init__(self, reduce: Optional[Callable[[JaxArray], JaxArray]]):
        self._reduce = reduce

    @property
    @abc.abstractmethod
    def value(self) -> JaxArray:
        raise NotImplementedError('Pure method')

    @value.setter
    @abc.abstractmethod
    def value(self, tensor: JaxArray):
        raise NotImplementedError('Pure method')

    def assign(self, tensor: JaxArray, check=True):
        # ...

    def reduce(self, tensors: JaxArray):
        # ...

    def __repr__(self):
        # ...


# ... definition of other classses 


# In the end of the file
# Support for JAX duck typing
_UNARY_OPS = [
    '__neg__',
    #...
]
_BINARY_OPS = [
    '__add__',
    #...
]

for unary_op in _UNARY_OPS:
    setattr(BaseVar, unary_op, lambda self, other, op_name=unary_op: getattr(self.value, op_name)())

for binary_op in BINARY_OPS:
    setattr(BaseVar, binary_op, lambda self, other, op_name=binary_op: getattr(self.value, op_name)(get_jax_value(other)))
```


This would remove clutter from `BaseVar` code, however will add clutter in the end of file and potentially may create confusion about how `BaseVar` works.

So I would be interested to hear opinions about it

aterzis-google (approved):

PR #209, opened 2021-03-19
Added description of how to to access RandomState in JIT-ted functions.

aterzis-google:


docs/source/advanced/jit.rst
```diff
@@ -206,6 +206,29 @@ A simple constant optimization example::
...
+otherwise the :code:`DEFAULT_GENERATOR` will be treated as a constant and the jitted function
+will always return the same value.
+
+.. code-block:: python
+
+    import objax
+    objax.random.DEFAULT_GENERATOR.seed(123)
+
+
+    def function():
```
AlexeyKurakin, line 222: Recommended style is use `Function.with_vars` instead of passing variable collection to JIT:

```python
@objax.Function.with_vars(objax.random.DEFAULT_GENERATOR.vars())
def function():
    d = objax.random.uniform((1,))
    return d

function_jit = objax.Jit(function)
```

docs/source/advanced/jit.rst
```diff
@@ -206,6 +206,29 @@ A simple constant optimization example::
     The XLA backend (the interface to the hardware) will do the constant optimization and may take a long time and
     a lot of memory due to compilation, often with very little gain in final performance, if any.
 
+RandomState and Jit
+^^^^^^^^^^^^^^^^^^^
+
+You need to pass :code:`objax.random.DEFAULT_GENERATOR` as a variable to :code:`VarCollection`,
```
AlexeyKurakin, line 212: User may use custom generator instead of default, we should mention this as well.

AlexeyKurakin (changes requested):
In addition to update of RST docs it's probably worth to also update docstring comments of each random function mentioning the need of passing generator to Jit.
Something like following:


```python
# inside objax/random/random.py
def normal(shape: Tuple[int, ...], *, mean: float = 0, stddev: float = 1, generator: Generator = DEFAULT_GENERATOR):
    """Returns a ``JaxArray`` of shape ``shape`` with random numbers from a normal distribution
    with mean ``mean`` and standard deviation ``stddev``.

    NOTE: if random numbers are generates inside jitted, parallelized or vectorized function then generator variables (including DEFAULT_GENERATOR)
    has to be added to variable collection."""
    return jr.normal(generator(), shape=shape) * stddev + mean
```

People may not read RST guide, but very likely people will check API docstring (either by looking at the code or via IDE tooltip) before using the API. Thus putting comment into docstring has higher chance of being seen by user.

docs/source/advanced/jit.rst
aterzis-google: updated

docs/source/advanced/jit.rst
aterzis-google: updated

aterzis-google:
Look at https://github.com/google/objax/pull/210 instead

PR #215, opened 2021-04-07
Remove requirement for file if pretrained is set to false for VGG19.

shs037:


AlexeyKurakin:
Could you please add unit test for VGG network?

PR #218, opened 2021-04-21
Update documentation to reflect that we don't need to use .value

aterzis-google:
anymore.

### DIFF
--- a/docs/source/advanced/variables_and_modules.rst
+++ b/docs/source/advanced/variables_and_modules.rst
@@ -188,8 +188,26 @@ Here's a basic example of RandomState manipulation::
 Accessing a variable's value
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-You do not have to use :code:`.value` to access the value of an Objax variable when using that variable in a computation.
+While you can use :code:`variable.value` to access a variable's value it can make the code noisy. Instead, when
+using variable :code:`v` in math expressions you can omit :code:`.value` as the example below demonstrates::
 
+    import objax
+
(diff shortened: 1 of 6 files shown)


docs/source/advanced/variables_and_modules.rst
```diff
@@ -185,6 +185,12 @@ Here's a basic example of RandomState manipulation::
     print(v.split(2))  # [[ 622232657  209145368] [2741198523 2127103341]]
     print(v.value)     # [3514448473 2078537737]
 
+Accessing a variable's value
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+You do not have to use :code:`.value` to access the value of an Objax variable when using that variable in a computation.
```
AlexeyKurakin, line 191: Maybe add an example here to clarify what this phrase mean:

```python
batch_size = 4
ndim = 16

w = objax.TrainVar(objax.random.normal((ndim,)))
b = objax.TrainVar(objax.random.normal((1,)))

x = objax.random.normal((batch_size, ndim))

# Both of the following statements work the same way:
y1 = jn.dot(x, w.value) + b.value
y2 = jn.dot(x, w) + b

print(jn.linalg.norm(y1 - y2))  # 0.0
```

Also add a phrase saying that if user need to update value of the variable then either `var.assign(new_value)` or `var.value = new_value` should be used.

aterzis-google: Added the suggested code snippet and text about using v.assign() to update a variable's value.

AlexeyKurakin (approved):

Issue #190, opened 2021-01-10
Raise warning when arguments to loss functions are not of the same dimensions


aterzis-google:
Loss functions (https://objax.readthedocs.io/en/latest/objax/functional.html?highlight=loss#objax-functional-loss) take two arguments which should have the same dimensions. However the functions do not check whether the dimensions are the same and when they are not they calculate something other than what the user expects. 

For example, consider the following function:

def loss(x, label):
  y_hat = model(x)[:, 0]  # This is the right one
  y_hat = model(x)   # This will not work
  return objax.functional.loss.mean_squared_error(y_hat, label, keep_axis=None)

When called with:

label.shape: (1000,)
model(x): (1000,1) 

mean_squared_error will not calculate what the user expects unless the user explicitly reshapes the input arguments.

PR #223, opened 2021-06-12
Raised WARNINGS if input arguments dont have same shape 

iamharsha1999:
Related to Issue #190. I have included warning checkpoints only to those functions that doesn't raise error by default in order to avoid unnecessary warnings.

### DIFF
--- a/objax/functional/loss.py
+++ b/objax/functional/loss.py
@@ -27,7 +27,7 @@ from objax.typing import JaxArray
 import warnings
 
 
-warn_mess = 'Argument shape mismatch in'
+WARN_SHAPE_MISMATCH = 'Argument shape mismatch in'
 
 
 def cross_entropy_logits(logits: JaxArray, labels: JaxArray) -> JaxArray:
@@ -40,8 +40,10 @@ def cross_entropy_logits(logits: JaxArray, labels: JaxArray) -> JaxArray:
     Returns:
         (batch, ...) tensor of the cross-entropies for each entry.
     """
+
+    FUNC_NAME = 'cross_entropy_logits'
     if logits.shape != labels.shape:
-        warnings.warn(' {} {} : arg1 {} and arg2 {}'.format(warn_mess, __all__[0], logits.shape, labels.shape))
+        warnings.warn(' {} {} : arg1 {} and arg2 {}'.format(WARN_SHAPE_MISMATCH, FUNC_NAME, logits.shape, labels.shape))
     return logsumexp(logits, axis=-1) - (logits * labels).sum(-1)
 
 
@@ -73,6 +75,7 @@ def l2(x: JaxArray) -> JaxArray:
     Returns:
         scalar tensor containing the l2 loss of x.
     """
+
     return 0.5 * (x ** 2).sum()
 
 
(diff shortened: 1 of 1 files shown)


iamharsha1999:
@googlebot I signed it!

iamharsha1999:
I have rectified the errors from linting and unit tests. Kindly verify and the merge.

iamharsha1999:
Can I know why its taking so long for workflow approval ?

iamharsha1999:
@david-berthelot @AlexeyKurakin Can you please approve the workflow and run the tests ?

objax/functional/loss.py
```diff
@@ -38,6 +39,9 @@ def cross_entropy_logits(logits: JaxArray, labels: JaxArray) -> JaxArray:
     Returns:
         (batch, ...) tensor of the cross-entropies for each entry.
     """
+    if logits.shape != labels.shape:
+        warnings.warn('Size Mismatch: logits {} and labels {}'.format(logits.shape, labels.shape))
```
AlexeyKurakin, line 43: Could you please add function name into each warning?
Like following: `'Argument shape mismatch in cross_entropy_logits: logits {} and labels {}'.format(logits.shape, labels.shape)'`

AlexeyKurakin (commented):
Sorry for delayed review and thanks for contributing.

The change looks good overall. I would just ask you to add function name into each of the warning. I provided example in the code comments. This would help for user to actually see where warning is happening, instead of guessing.

objax/functional/loss.py
iamharsha1999: Hey @AlexeyKurakin have made the following changes as per your comment. I added a separate decorator for this purpose in order to keep the code simple. Kindly review the changes and merge.

AlexeyKurakin: I would actually ask to use simple warnings instead of decorator. Generally decorators in Python adds a level of indirection, thus understanding the code will require looking at the code of the decorator.
In Objax we prefer to have explicit code whenever possible. This makes it easier to know what code is doing and easier to debug code.  As far I as remember, the only place where we use and encourage decorator is `objax.Function.with_vars` or `objax.Function.auto_vars`.

Also I would ask to avoid overriding warnings formatter. User of the library may have their own customization of warning formatter, which will interfere with overrides done in Objax.

iamharsha1999: Ohh then would change it to the old way. No issues.

objax/functional/loss.py
```diff
@@ -38,6 +40,8 @@ def cross_entropy_logits(logits: JaxArray, labels: JaxArray) -> JaxArray:
     Returns:
         (batch, ...) tensor of the cross-entropies for each entry.
     """
+    if logits.shape != labels.shape:
+        warnings.warn(' {} {} : arg1 {} and arg2 {}'.format(warn_mess, __all__[0], logits.shape, labels.shape))
```
AlexeyKurakin, line 44: Using `__all__[index]` is very brittle to refer to function name.
If somebody later change order of functions in `__all__` then all warnings will be broken.
The simple way is to just explicitly write function name as a string: `warningns.warn(' .... '. format(WARN_MESS, 'cross_entropy_logits', logits.shape, labels.shape)`

You can look into programmatically getting function name from within a function.
The best option I found is `inspect.currentframe().f_code.co_name`, but it's not particularly nice or readable.
See here for some discussion: https://stackoverflow.com/questions/5067604/determine-function-name-from-within-that-function-without-using-traceback

objax/functional/loss.py
```diff
@@ -21,11 +21,13 @@
...
 from typing import Union, Iterable, Optional
-
 import jax.numpy as jn
-
 from objax.functional.core import logsumexp
 from objax.typing import JaxArray
+import warnings
+
+
+warn_mess = 'Argument shape mismatch in'
```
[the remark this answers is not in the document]
iamharsha1999: The main reason I avoided to use the function names explicitly was to keep the code length lesser than 120 characters per line. Otherwise the linter returns an error. Now I have added an another constant `FUNC_NAME` in each functions definition and then passed it to the warning message. I guess this one should be perfect. Kindly do verify.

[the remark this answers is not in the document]
AlexeyKurakin: Yes, that's fine.
FYI, if line is too long you can split it into multiple lines.

AlexeyKurakin (approved):

PR #242, opened 2022-08-05
Update default values for batchnorm

yechengxi:
change default batch norm momentum and eps to match the pytorch implementation. The previous setting gave us a slow update speed.


aterzis-google:
Thanks for the PR Chengxi!

It's not one of Objax's goals to mirror PyTorch so it's OK to have different defaults. Also we would like to see more evidence that the new default values are better across a wide variety of settings.

PR #247, opened 2022-08-26
Fix test errors and some warnings

aterzis-google:


tests/sequential.py
```diff
@@ -108,18 +108,19 @@ def __call__(self, x, y):
     def test_on_sequential_missing_argument(self):
         m = objax.nn.Sequential([objax.nn.Linear(2, 3), objax.nn.BatchNorm0D(3), objax.nn.Linear(3, 2)])
         x = jn.array([[1., -1.], [2., -2.]])
-        msg = f'Sequential layer[1] {m[1]} __call__() missing 1 required positional argument: \'training\''
+        error_substring = "missing 1 required positional argument: 'training'"
+        msg = f'{error_substring}'
```
AlexeyKurakin, line 112: I think you don't actually two variables `msg` and `error_substring` with the same value.
Just keep only one

tests/sequential.py
```diff
@@ -108,18 +108,19 @@ def __call__(self, x, y):
...
         try:
             m(x)
             assert False
         except TypeError as e:
-            self.assertEqual(str(e), msg)
+            self.assertTrue(str(e).find(msg) != -1)
```
AlexeyKurakin, line 117: I wonder if you can use `self.assertIn(msg, str(e))` instead?
This would be more readable. In python `a in b` for strings returns True when `a` is a substring of `b` and False otherwise. So I think `assertIn` might be appropriate here

AlexeyKurakin (approved):

PR #250, opened 2022-11-16
Update nn.rst

divyas248:
The channel number for 'in' is currently set as c which is incorrect because c is referring to the output channel number. Instead this needs to be set as t (which is the variable that iterates over the input channel numbers). 
in[n,c,i+h,j+w] should be changed to in[n,t,i+h,j+w]


aterzis-google (commented):
Thank you for submitting this change and apologies it took us so long to get back to you!

An equivalent change was submitted in https://github.com/google/objax/pull/252/ which we merged.

PR #266, opened 2023-11-01
Fix tests following jax update

GaetanLepage:
Some tests were not adapted to the recent changes in the jax API.
This PR ensures that all tests pass with the most recent jax.


aterzis-google:
Gaetan thank you for the PR to fix the failing tests!

I just merged a PR that is equivalent so we won't be needing this change. 

Thanks again!

GaetanLepage:
Ok ! Thank you :)

